PR #10, opened 2019-12-28
CI

TheGlitch76:
Adds everything the other PRs did, plus publishing to Github Releases on tag.
However (like the other PRs because I missed a couple things i.e. fabric.mod.json) versions must be set manually instead of being determined by git magic, mostly because I'm scared to touch things that aren't broken.

Draft because the test mod support needs to be re-added.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,36 @@
+name: Java CI
+
+on:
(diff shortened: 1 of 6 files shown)


TheGlitch76:
Ready for review + squash merge

PR #27, opened 2020-01-17
Jar-In-Jar EventBus

shedaniel:


build.gradle
```diff
@@ -230,6 +230,7 @@ repositories {
 		dirs 'jars'
 	}
 	mavenCentral()
+	maven { url 'https://dl.bintray.com/patchworkmc/Patchwork-Maven/' }
```
natanfudge: Click the "upload to jcenter" button so you can just put `jcenter()` and that will include both patchwork and maven central, which is cleaner and improves build performance

shedaniel: I currently have no access to this bintray, we need to ask coderbot.

natanfudge: @coderbot16

TheGlitch76: getting approved takes a couple business days, i think this should be merged now and deal with that later

PR #32, opened 2020-01-20
ObfuscationReflectionHelper

TheGlitch76:
This was sitting in my git stash and I forgot it existed. I'll have to go through and test again to make sure it works, but reviews are welcome now.


patchwork-fml/src/main/java/cpw/mods/modlauncher/api/INameMappingService.java
```diff
@@ -0,0 +1,27 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package cpw.mods.modlauncher.api;
+
+/**
+ * Only serves to expose the Domain enum.
+**/
```
coderbot16:
```suggestion
 */
```

patchwork-fml/src/main/java/cpw/mods/modlauncher/api/INameMappingService.java
```diff
@@ -0,0 +1,27 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package cpw.mods.modlauncher.api;
```
coderbot16: Could this cause issues if Fabric is launched using ModLauncher? I would assume the true class would get loaded in that case, however.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+	 * @return The method with the specified name and parameters in the given class.
+	 * @throws NullPointerException        If {@code clazz} is null.
+	 * @throws NullPointerException        If {@code methodName} is null.
+	 * @throws IllegalArgumentException    If {@code methodName} is empty.
+	 * @throws NullPointerException        If {@code parameterTypes} is null.
+	 * @throws UnableToFindMethodException If the method could not be found.
+	 */
+	@Nonnull
+	public static Method findMethod(@Nonnull final Class<?> clazz, @Nonnull final String methodName, @Nonnull final Class<?>... parameterTypes) {
+		Preconditions.checkNotNull(clazz, "Class to find method on cannot be null.");
```
coderbot16: Can't this be replaced with `Objects.requireNonNull`?

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+		Preconditions.checkNotNull(methodName, "Name of method to find cannot be null.");
+		Preconditions.checkArgument(!methodName.isEmpty(), "Name of method to find cannot be empty.");
+		Preconditions.checkNotNull(parameterTypes, "Parameter types of method to find cannot be null.");
+
+		try {
+			Method m = clazz.getDeclaredMethod(PatchworkMappingResolver.remapName(INameMappingService.Domain.METHOD, clazz, methodName), parameterTypes);
```
coderbot16: This line seems a little long, maybe split it up? Maybe `method` instead of `m` too.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+	 * @param parameterTypes The parameter types of the constructor.
+	 * @param <T>            The type.
+	 * @return The constructor with the specified parameters in the given class.
+	 * @throws NullPointerException        If {@code clazz} is null.
+	 * @throws NullPointerException        If {@code parameterTypes} is null.
+	 * @throws UnknownConstructorException If the constructor could not be found.
+	 */
+	@Nonnull
+	public static <T> Constructor<T> findConstructor(@Nonnull final Class<T> clazz, @Nonnull final Class<?>... parameterTypes) {
+		Preconditions.checkNotNull(clazz, "Class to find constructor on cannot be null.");
```
coderbot16: `Objects.requireNonNull` again

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+		} catch (Exception e) {
+			throw new UnableToFindMethodException(e);
+		}
+	}
+
+	/**
+	 * Finds a constructor with the specified parameter types in the given class and makes it accessible.
+	 * Note: For performance, store the returned value and avoid calling this repeatedly.
+	 *
+	 * <p>Throws an exception if the constructor is not found.</p>
```
coderbot16:
```suggestion
	 * <p>Throws an exception if the constructor was not found.</p>
```

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+	 * @throws UnableToFindFieldException If the field could not be found.
+	 */
+	@Nonnull
+	public static <T> Field findField(@Nonnull final Class<? super T> clazz, @Nonnull final String fieldName) {
+		Preconditions.checkNotNull(clazz, "Class to find field on cannot be null.");
+		Preconditions.checkNotNull(fieldName, "Name of field to find cannot be null.");
+		Preconditions.checkArgument(!fieldName.isEmpty(), "Name of field to find cannot be empty.");
+
+		try {
+			Field f = clazz.getDeclaredField(PatchworkMappingResolver.remapName(INameMappingService.Domain.FIELD, clazz, fieldName));
```
coderbot16: This line seems a little long too; maybe `field` instead?

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+
+import com.patchworkmc.impl.fml.PatchworkMappingResolver;
+
+/**
+ * Some reflection helper code.
+ * This may not work properly in Java 9 with its new, more restrictive, reflection management.
+ * As such, if issues are encountered, please report them and we can see what we can do to expand
+ * the compatibility.
+ *
+ * <p>In other cases, AccessTransformers may be used.</p>
```
coderbot16: Mixins? This comment isn't super relevant to Patchwork, since ATs don't exist at runtime (unless we're talking about FabricASM)

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * <p>All field and method names should be passed in as intermediary names, and this will automatically resolve if Yarn mappings are detected.</p>
+ *
+ */
+@SuppressWarnings({ "serial", "unchecked", "unused", "WeakerAccess" })
```
coderbot16: These warnings are scary, if they're only relevant for a few lines maybe move them to a more restricted scope?

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,340 @@
...
+	 * @deprecated Use {@link #getPrivateValue(Class, Object, String)} because field indices change a lot more often than field names do.
+	 * Forge: TODO Remove in 1.15
+	 */
+	@Deprecated
+	@Nullable
+	public static <T, E> T getPrivateValue(Class<? super E> classToAccess, E instance, int fieldIndex) {
+		try {
+			return (T) findField(classToAccess, fieldIndex).get(instance);
+		} catch (Exception e) {
+			LOGGER.error(REFLECTION, "There was a problem getting field index {} from {}", fieldIndex, classToAccess.getName(), e);
```
coderbot16: Does this actually print the exception?

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingResolver.java
```diff
@@ -0,0 +1,140 @@
...
+			if (c.isArray()) {
+				return className.replace('.', '/');
+			} else {
+				return ('L' + className + ';').replace('.', '/');
+			}
+		}
+	}
+
+	private static String getMethodDescriptor(Method m, boolean remapToIntermediary) {
+		StringBuilder sb = new StringBuilder();
```
coderbot16: `descriptor`?

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingResolver.java
```diff
@@ -0,0 +1,140 @@
...
+
+			if (c.isArray()) {
+				return className.replace('.', '/');
+			} else {
+				return ('L' + className + ';').replace('.', '/');
+			}
+		}
+	}
+
+	private static String getMethodDescriptor(Method m, boolean remapToIntermediary) {
```
coderbot16:
```suggestion
	private static String getMethodDescriptor(Method method, boolean remapToIntermediary) {
```

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingResolver.java
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			return resolver.mapFieldName(NAMED, className, name, fieldDescriptor);
+		default:
+			throw new IllegalArgumentException("Someone's been tampering with enums! Got unexpected type " + domain.name());
+		}
+	}
+
+	// this is a stackoverflow response but with remapping
+	private static String getDescriptorForClass(final Class<?> c, boolean remapToIntermediary) {
```
coderbot16:
```suggestion
	private static String getDescriptorForClass(final Class<?> clazz, boolean remapToIntermediary) {
```

patchwork-fml/src/main/java/cpw/mods/modlauncher/api/INameMappingService.java
TheGlitch76: Based on a bit of googling--as long as ModLauncher is loaded before Patchwork (which is true) it *should* just not load our stripped class.
The only way to know for absolutely sure is to test, but it should work fine theoretically.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
TheGlitch76: Technically, yes, but IMO it's better for consistency because methods that have no equivalent in Objects like `Preconditions.checkArgument` are used elsewhere in this class.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
TheGlitch76: Yeah, it will print the exception and its stacktrace on a new line. I tested it

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
TheGlitch76: See above

TheGlitch76:
This seems to be working fine.

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingResolver.java
```diff
@@ -0,0 +1,140 @@
...
+			for (Method method : clazz.getDeclaredMethods()) {
+				// If the field name remapped to intermediary is the one we're looking for, store its descriptor and return
+				if (resolver.mapMethodName(INTERMEDIARY, className, method.getName(),
+						getMethodDescriptor(method, false)).equals(name)) {
+					// We need to remap the descriptor to intermediary in case it contains named Minecraft classes
+					methodDescriptor = getMethodDescriptor(method, true);
+					break;
+				}
+			}
+
```
coderbot16: Should probably null-check here

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingResolver.java
```diff
@@ -0,0 +1,140 @@
...
+			for (Field field : clazz.getDeclaredFields()) {
+				// If the field name remapped to intermediary is the one we're looking for, store its descriptor and return
+				if (resolver.mapFieldName(INTERMEDIARY, className, field.getName(),
+						getDescriptorForClass(field.getType(), false)).equals(name)) {
+					// We need to remap the descriptor to intermediary in case it contains named Minecraft classes
+					fieldDescriptor = getDescriptorForClass(field.getType(), true);
+					break;
+				}
+			}
+
```
coderbot16: Null check here too

TheGlitch76:
It does not work fine, I just messed up my tests.

TheGlitch76:
Works as intended now, as far as I can tell.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+ * <p>All field and method names should be passed in as intermediary names, and this will automatically resolve if Yarn mappings are detected.</p>
+ *
+ */
+public class ObfuscationReflectionHelper {
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final Marker REFLECTION = MarkerManager.getMarker("REFLECTION");
+	private static final TinyTree MAPPINGS = FabricLauncherBase.getLauncher().getMappingConfiguration().getMappings();
+	private static final String INTERMEDIARY = "intermediary";
+	private static final String NAMED = "named";
+	/**
```
coderbot16: Blank lines before documentation comments
```suggestion

	/**
```

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+/**
+ * Some reflection helper code.
+ * This may not work properly in Java 9 with its new, more restrictive, reflection management.
+ * As such, if issues are encountered, please report them and we can see what we can do to expand
+ * the compatibility.
+ *
+ * <p>In other cases, accesssor mixins may be used.</p>
+ *
+ * <p>All field and method names should be passed in as intermediary names, and this will automatically resolve if Yarn mappings are detected.</p>
+ *
```
coderbot16:
```suggestion
```

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+	/**
+	 * Remaps a name from intermediary to whatever is currently being used at runtime.
+	 *
+	 * @param domain The {@link INameMappingService.Domain} to use to remap the name.
+	 * @param name   The name to try and remap.
+	 * @return The remapped name, or the original name if it couldn't be remapped.
+	 */
+	@Nonnull
+	public static String remapName(INameMappingService.Domain domain, String name) {
+		if (!FabricLoader.getInstance().isDevelopmentEnvironment()) {
```
coderbot16: It could make more sense just to check for intermediary, there's no guarantee that someone won't show up with some yarn-at-runtime mod or something. We're remapping from intermediary anyways.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+ *
+ * <p>In other cases, accesssor mixins may be used.</p>
+ *
+ * <p>All field and method names should be passed in as intermediary names, and this will automatically resolve if Yarn mappings are detected.</p>
+ *
+ */
+public class ObfuscationReflectionHelper {
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final Marker REFLECTION = MarkerManager.getMarker("REFLECTION");
+	private static final TinyTree MAPPINGS = FabricLauncherBase.getLauncher().getMappingConfiguration().getMappings();
```
coderbot16: I feel that we should at least put some note here about the fact that we're reaching around in internal Fabric APIs, maybe a TODO or something.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+	 */
+	@Nonnull
+	public static Method findMethod(@Nonnull final Class<?> clazz, @Nonnull final String methodName, @Nonnull final Class<?>... parameterTypes) {
+		Preconditions.checkNotNull(clazz, "Class to find method on cannot be null.");
+		Preconditions.checkNotNull(methodName, "Name of method to find cannot be null.");
+		Preconditions.checkArgument(!methodName.isEmpty(), "Name of method to find cannot be empty.");
+		Preconditions.checkNotNull(parameterTypes, "Parameter types of method to find cannot be null.");
+
+		try {
+			String name = remapName(INameMappingService.Domain.METHOD, methodName);
```
coderbot16: We already know the class here, I don't think we should scan over every single method def in every single class instead of just looking through that class directly.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+	 * @throws UnableToFindFieldException If the field could not be found.
+	 */
+	@Nonnull
+	public static <T> Field findField(@Nonnull final Class<? super T> clazz, @Nonnull final String fieldName) {
+		Preconditions.checkNotNull(clazz, "Class to find field on cannot be null.");
+		Preconditions.checkNotNull(fieldName, "Name of field to find cannot be null.");
+		Preconditions.checkArgument(!fieldName.isEmpty(), "Name of field to find cannot be empty.");
+
+		try {
+			String name = remapName(INameMappingService.Domain.FIELD, fieldName);
```
coderbot16: Same here, but for fields instead.

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,375 @@
...
+	 * @throws UnableToAccessFieldException If there was a problem getting the value.
+	 */
+	@Nullable
+	public static <T, E> T getPrivateValue(Class<? super E> classToAccess, E instance, String fieldName) {
+		try {
+			//noinspection unchecked
+			return (T) findField(classToAccess, fieldName).get(instance);
+		} catch (UnableToFindFieldException e) {
+			LOGGER.error(REFLECTION, "Unable to locate field {} ({}) on type {}", fieldName,
+					remapName(INameMappingService.Domain.FIELD, fieldName), classToAccess.getName(), e);
```
coderbot16: I don't really like looking up the name here again, even if it's just on the error path...

TheGlitch76: We actually have to do the expensive lookup instead of the fast one, because this only throws when the field isn't found. Not doing the expensive lookup would reduce error readability in the event that someone looked up the right field name in the wrong class.

TheGlitch76:
I'd like to handle some deduplication and then I'll be good to go.

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingService.java
```diff
@@ -0,0 +1,114 @@
...
+import cpw.mods.modlauncher.api.INameMappingService;
+
+import net.fabricmc.loader.api.FabricLoader;
+import net.fabricmc.loader.launch.common.FabricLauncherBase;
+import net.fabricmc.mapping.tree.ClassDef;
+import net.fabricmc.mapping.tree.Mapped;
+import net.fabricmc.mapping.tree.TinyTree;
+
+public class PatchworkMappingService {
+	public static final TinyTree MAPPINGS = FabricLauncherBase.getLauncher().getMappingConfiguration().getMappings();
```
coderbot16: These should be private

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingService.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Like {@link PatchworkMappingService#remapNameFast(INameMappingService.Domain, Class, String)}, but takes a {@link ClassDef} instead of a {@link Class}.
+	 * @param domain    The {@link INameMappingService.Domain} to look up.
+	 * @param classDef  The classDef that contains the {@code name} to look up.
+	 * @param name      The name to remap.
+	 * @return The remapped name, or null if it couldn't be remapped.
+	 */
+	@Nullable
+	public static String remapNameInternal(INameMappingService.Domain domain, ClassDef classDef, String name) {
```
coderbot16: This should be private

patchwork-fml/src/main/java/net/minecraftforge/fml/common/ObfuscationReflectionHelper.java
```diff
@@ -0,0 +1,338 @@
...
+ * the compatibility.
+ *
+ * <p>In other cases, accesssor mixins may be used.</p>
+ *
+ * <p>All field and method names should be passed in as intermediary names, and this will automatically resolve if Yarn mappings are detected.</p>
+ */
+public class ObfuscationReflectionHelper {
+	private static final Logger LOGGER = LogManager.getLogger();
+	private static final Marker REFLECTION = MarkerManager.getMarker("REFLECTION");
+	// We're technically messing with Loader's internal APIs here, if Loader ever gets a better mapping resolution system this class should be refactored.
```
coderbot16: This should get moved to PatchworkMappingService

patchwork-fml/src/main/java/com/patchworkmc/impl/fml/PatchworkMappingService.java
```diff
@@ -0,0 +1,114 @@
...
+	/**
+	 * Like {@link PatchworkMappingService#remapName(INameMappingService.Domain, String)}, but only iterates through members of the target class.
+	 * @param domain    The {@link INameMappingService.Domain} to look up.
+	 * @param clazz     The class that contains the {@code name} to look up.
+	 * @param name      The name to remap.
+	 * @return The remapped name, or the original name if it couldn't be remapped.
+	 */
+	@Nonnull
+	public static String remapNameFast(INameMappingService.Domain domain, Class<?> clazz, String name) {
+		ClassDef classDef = MAPPINGS.getDefaultNamespaceClassMap().get(clazz.getName());
```
coderbot16: This should have the intermediary check too (avoid executing any remap code in prod), and also I think we should deduplicate it into a private helper function (`isIntermediary` or something similar?)

PR #50, opened 2020-03-03
keybinds

shedaniel:
welcome to the most controversial module ever!1!!1!11
this pr will edit vanilla behaviour, is that what we are avoiding? it might be hard to avoid all of that.
we can probably add this module but not include it in our main jar, but release it as a separate jar for users to have _options_.

this is only step 1, we still have to redirect classes using KeyBinding to PatchworkKeyBinding.

Signed-off-by: shedaniel <daniel@shedaniel.me>


patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+		for (KeyBinding keybinding : ((KeyBindingMap) keysByCode).lookupAll(keyCode)) {
+			if (keybinding != null) {
+				((KeyBindingHooks) keybinding).setPressed(pressed);
+			}
+		}
+	}
+
+	@Inject(method = "isPressed", at = @At("RETURN"), cancellable = true)
+	private void isPressed(CallbackInfoReturnable<Boolean> cir) {
+		cir.setReturnValue(cir.getReturnValueZ() && getKeyConflictContext().isActive() && getKeyModifier().isActive(getKeyConflictContext()));
```
coderbot16:
```suggestion
		KeyConflictContext context = getKeyConflictContext();
		
		if (!context.isActive() || !getKeyModifier().isActive(context)) {
			cir.setReturnValue(false);
		}
```

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+	@Shadow
+	@Final
+	private static Map<String, Integer> categoryOrderMap;
+	private KeyModifier keyModifierDefault = KeyModifier.NONE;
+	private KeyModifier keyModifier = KeyModifier.NONE;
+	private IKeyConflictContext keyConflictContext = KeyConflictContext.UNIVERSAL;
+
+	@Inject(method = "<clinit>", at = @At("RETURN"))
+	private static void staticInit(CallbackInfo ci) {
+		keysByCode = new KeyBindingMap();
```
coderbot16: I dislike this approach of overwriting the vanilla map and behavior. Instead, it makes more sense to just have a secondary map for all of the non-NONE KeyModifiers alongside the vanilla map, and then in the methods that reference the vanilla map update the modifer map as well.

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	@Inject(method = "onKeyPressed", at = @At("HEAD"), cancellable = true)
+	private static void onKeyPressed(InputUtil.KeyCode keyCode, CallbackInfo ci) {
+		ci.cancel();
```
coderbot16: This is a silent `@Overwrite`, see my comment above.

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+		for (KeyBinding keybinding : ((KeyBindingMap) keysByCode).lookupAll(keyCode)) {
+			if (keybinding != null) {
+				((KeyBindingHooks) keybinding).setTimesPressed(((KeyBindingHooks) keybinding).getTimesPressed() + 1);
+			}
+		}
+	}
+
+	@Inject(method = "setKeyPressed", at = @At("HEAD"), cancellable = true)
+	private static void onKeyPressed(InputUtil.KeyCode keyCode, boolean pressed, CallbackInfo ci) {
+		ci.cancel();
```
coderbot16: This is a silent `@Overwrite`, see my comment above.

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+
+	@Inject(method = "isPressed", at = @At("RETURN"), cancellable = true)
+	private void isPressed(CallbackInfoReturnable<Boolean> cir) {
+		cir.setReturnValue(cir.getReturnValueZ() && getKeyConflictContext().isActive() && getKeyModifier().isActive(getKeyConflictContext()));
+	}
+
+	@Inject(method = "method_1430", at = @At("HEAD"), cancellable = true)
+	private void compareTo(KeyBinding keyBinding, CallbackInfoReturnable<Integer> cir) {
```
coderbot16: After looking at the implementation in Forge, it looks like the purpose of this patch is to handle null entries in the categoryOrderMap, no more. This injector is a silent `@Overwrite`, I think this should be rewritten to just fulfill the purpose of handling nulls without clobbering vanilla behavior.
```suggestion
	private void handleNullCategories(KeyBinding keyBinding, CallbackInfoReturnable<Integer> cir) {
```

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+		if (tCat == null && oCat == null) {
+			cir.setReturnValue(I18n.translate(this.getCategory()).compareTo(I18n.translate(keyBinding.getCategory())));
+			return;
+		}
+
+		cir.setReturnValue(tCat.compareTo(oCat));
+	}
+
+	@Inject(method = "equals", at = @At("HEAD"), cancellable = true)
+	private void equals(KeyBinding binding, CallbackInfoReturnable<Boolean> cir) {
```
coderbot16: This function has a lot of repetition of certain fragments of code (such as `((IForgeKeybinding) binding).getKeyConflictContext()`), reducing the line count but making it harder to read. While Forge wants to reduce line count in patches, I think it is best to expand that kind of stuff here.

```suggestion
	private void conflictAwareEquals(KeyBinding binding, CallbackInfoReturnable<Boolean> cir) {
```

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+				cir.setReturnValue(keyModifier == otherKeyModifier
+								|| (getKeyConflictContext().conflicts(KeyConflictContext.IN_GAME)
+								&& (keyModifier == KeyModifier.NONE || otherKeyModifier == KeyModifier.NONE)));
+			}
+		}
+	}
+
+	@Inject(method = "getLocalizedName()Ljava/lang/String;", at = @At("RETURN"), cancellable = true)
+	private void getLocalizedName(CallbackInfoReturnable<String> cir) {
+		cir.setReturnValue(getKeyModifier().getLocalizedComboName(keyCode, cir::getReturnValue));
```
coderbot16: This should only cancel if the modifier is not NONE to avoid messing with other mods.

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	@Inject(method = "isDefault", at = @At("RETURN"), cancellable = true)
+	private void isDefault(CallbackInfoReturnable<Boolean> cir) {
+		cir.setReturnValue(cir.getReturnValueZ() && getKeyModifier() == getKeyModifierDefault());
```
coderbot16: In these cases it is better to encode the short-circuiting into your callback. Otherwise, you are potentially breaking every other RETURN or TAIL injector, since cancelling the function will prevent them from running after your injector.

In general, it is probably a mistake if your injector unconditionally uses a `setReturnValue`, since it's likely that you're just making an `@Overwrite` in disguise.

```suggestion
		if (getKeyModifier() != getKeyModifierDefault() {
			cir.setReturnValue(false);
		}
```

patchwork-extensions-keybinds/src/main/java/net/minecraftforge/client/extensions/IForgeKeybinding.java
```diff
@@ -0,0 +1,69 @@
...
+	KeyModifier getKeyModifier();
+
+	void setKeyModifierAndCode(KeyModifier keyModifier, InputUtil.KeyCode keyCode);
+
+	/**
+	 * Returns true when one of the bindings' key codes conflicts with the other's modifier.
+	 */
+	default boolean hasKeyCodeModifierConflict(KeyBinding other) {
+		if (getKeyConflictContext().conflicts(((IForgeKeybinding) other).getKeyConflictContext()) || ((IForgeKeybinding) other).getKeyConflictContext().conflicts(getKeyConflictContext())) {
+			return getKeyModifier().matches(((IForgeKeybinding) other).getKey()) || ((IForgeKeybinding) other).getKeyModifier().matches(getKey());
```
coderbot16: l o n g lines

patchwork-extensions-keybinds/src/main/java/net/minecraftforge/client/settings/KeyBindingMap.java
```diff
@@ -0,0 +1,294 @@
...
+import com.google.common.collect.Sets;
+import net.minecraftforge.client.extensions.IForgeKeybinding;
+
+import net.minecraft.client.options.KeyBinding;
+import net.minecraft.client.util.InputUtil;
+
+/**
+ * This is implementing Map to add vanilla compatibility.
+ */
+public class KeyBindingMap implements Map<InputUtil.KeyCode, KeyBinding> {
```
coderbot16: See my comment in `MixinKeyBinding`.

patchwork-extensions-keybinds/src/main/java/net/minecraftforge/client/settings/KeyConflictContext.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package net.minecraftforge.client.settings;
+
+import net.minecraft.client.MinecraftClient;
+
+public enum KeyConflictContext implements IKeyConflictContext {
+	/**
+	 * Universal key bindings are used in every context and will conflict with any other context.
+	 * Key Bindings are universal by default.
```
coderbot16:
```suggestion
	 * Key bindings are universal by default.
```

patchwork-extensions-keybinds/src/main/java/com/patchworkmc/mixin/extensions/keybinds/MixinKeyBinding.java
shedaniel: Forge allows two keybinds to have the same keycode with this approach, vanilla is really bad at this.

Vanilla is just a map from KeyCode to KeyBinding.
```java
private static final Map<InputUtil.KeyCode, KeyBinding> keysByCode = Maps.newHashMap();
```
However, I don't think overwriting this is such a bad idea, this mimics the vanilla map function and anyone using the map would be compatible.

The next question is do we really have to support multiple keybinds with the same keycode, probably not, but I think it is kind of worth it if we are going to implement modifiers anyways, we are editing vanilla behaviour anyways.

coderbot16: What I'm trying to say is have the maps for each other key modifier exist alongside the vanilla map, because effectively the vanilla map is already working as the map for the NONE keymodifier so replacing it is unnecessary.

shedaniel: We still have to overwrite the methods if the user uses the modifers, and other mods using the map will simply ignore the modifiers.

PR #51, opened 2020-03-07
3 new entity events

SuperCoder7979:
Implement the rest of the entity events.


patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/EntityEvent.java
```diff
@@ -60,11 +62,11 @@ public Entity getEntity() {
...
-	/* TODO public static class EntityConstructing extends EntityEvent {
+	public static class EntityConstructing extends EntityEvent {
 		public EntityConstructing(Entity entity) {
 			super(entity);
 		}
-	}*/
+	}
 
 	/**
 	 * <b>TODO: Forge never fires this event.</b>
```
coderbot16: Looks like forge does fire this event

SuperCoder7979: concern

PR #59, opened 2020-03-24
Implement Annotation Scanning

qouteall:


patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,192 @@
...
+			if (annotationData == null) {
+				initAnnotationData();
+			}
+
+			return annotationData;
+		}
+
+		private void initAnnotationData() {
+			try {
+				Class<?> clazzObj = Class.forName(clazz.getClassName());
```
coderbot16:
```suggestion
				// TODO: This *may* load classes in the wrong order, but it shouldn't be an issue
				Class<?> clazzObj = Class.forName(clazz.getClassName());
```

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,192 @@
...
+		private Annotation getAnnotationObject(Class<?> clazzObj, Class annotationType) {
+			try {
+				switch (targetType) {
+				case TYPE:
+					return clazzObj.getAnnotation(annotationType);
+				case FIELD:
+					return clazzObj.getField(memberName)
+							.getAnnotation(annotationType);
+				case METHOD:
+					System.out.println("Not being supported");
```
coderbot16: This should be pretty easy to support though?

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,192 @@
...
+					return null;
+				default:
+					return null;
+				}
+			} catch (NoSuchFieldException e) {
+				return null;
```
coderbot16: This will immediately cause a NullPointerException in initAnnotationData

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
ramidzkh: This shouldn't load classes in the first place

coderbot16: Not doing so would require serializing the annotation properties into the JSON which could be complex. In fact, I suggested just reading it using reflection instead, which does have this issue. Though, as you've pointed out, that approach could be problematic.

PR #61, opened 2020-04-24
Add patchwork-recipes and implement IShapedRecipe

qouteall:


coderbot16:
The code looks good but I don't think we can merge this until the build passes.

patchwork-recipes/src/main/java/net/patchworkmc/mixin/recipes/MixinRecipeGridAligner.java
```diff
@@ -0,0 +1,99 @@
...
+	 */
+	@Overwrite
+	default void alignRecipeToGrid(
+			int gridWidth, int gridHeight, int gridOutputSlot,
+			Recipe<?> recipe, Iterator<?> inputs, int amount
+	) {
+		int width = gridWidth;
+		int height = gridHeight;
+
+		if (recipe instanceof IShapedRecipe) {
```
TheGlitch76, line 50: Can we get some markers on where the actual code changes are with comments? I'm assuming it's just here but it's difficult to tell.

coderbot16: Yes, it's generally good practice to annotate the changed parts when using `@Overwrite`.

PR #62, opened 2020-04-25
Annotation Scanning

qouteall:


patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	private void init() {
+		initialized = true;
+
+		ModContainer modContainer = getModContainer(modid);
+		CustomValue customValue = modContainer.getMetadata().getCustomValue("patchwork:annotations");
+
+		if (customValue == null) {
+			LOGGER.error("ModFileScanData is being accessed but cannot find annotation storage");
```
coderbot16, line 68: Should print the modid here for better debugging
```suggestion
			LOGGER.error("ModFileScanData: Tried to access the scanned annotation data for " + modid + ", but it is missing");
```

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,257 @@
...
+			));
+			e.printStackTrace();
+		}
+	}
+
+	private void initIfNeeded() {
+		if (!initialized) {
+			init();
+		}
+	}
```
coderbot16, line 97: Could probably just move this to getAnnotations since it's the only caller

patchwork-fml/src/main/java/net/minecraftforge/fml/ModList.java
```diff
@@ -19,12 +19,19 @@
...
 package net.minecraftforge.fml;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import net.minecraftforge.fml.loading.moddiscovery.ModFileInfo;
+
 import net.fabricmc.loader.api.FabricLoader;
 
 public class ModList {
```
coderbot16, line 29: Should probably have `getAllScanData`, somehow

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,257 @@
...
+				return false;
+			}
+
+			return jars.stream().anyMatch(
+					jar -> jar.getFile().contains(modid)
+			);
+		}).findAny().orElse(FabricLoader.INSTANCE.getModContainer(modid).orElseThrow(
+				() -> new RuntimeException("Cannot get mod container for " + modid)
+		));
+	}
```
coderbot16, line 256: I really don't like this. How about we have a custom attribute that designates the "owner" modid for the annotation data?

qouteall: I found that there is "modmenu:parent" in fabric.mod.json. Using that could be appropriate

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,250 @@
...
+
+	//if it's a jij mod, return parent mod's container
+	private static ModContainer getModContainer(String modid) {
+		ModContainer modContainer = FabricLoader.INSTANCE.getModContainer(modid)
+				.orElseThrow(() -> new RuntimeException("Cannot get mod container " + modid));
+
+		CustomValue customValue = modContainer.getMetadata().getCustomValue("modmenu:parent");
+
+		if (customValue == null) {
```
TheGlitch76, line 239: How does this prove that a mod is JIJ or not? Anybody can set a parent to any mod, and a JIJ mod does not have to have a parent.

qouteall: Normal Fabric mod does not need annotation scanning, and patched mods have this added by the patcher.

TheGlitch76: I'm not sure they do, but this would also match non-patchwork mods. It would be best to make a PR to patcher that adds a "patchwork:parent = foo" to the fabric mod json.

coderbot16: How does Forge deal with this? Surely it doesn't try to assign annotation data to a single mod ID because it's a property of the mod file. Maybe we need a patchwork:annotation_holder attribute to hold the target mod ID.

patchwork-fml/src/main/java/net/minecraftforge/fml/ModList.java
```diff
@@ -33,4 +45,20 @@ public boolean isLoaded(String modId) {
...
+	public ModFileInfo getModFileById(String modId) {
+		return modFileInfoMap.computeIfAbsent(modId, ModFileInfo::new);
+	}
+
+	public List<ModFileScanData> getAllScanData() {
+		if (allScanDataCache == null) {
+			allScanDataCache = FabricLoader.getInstance().getAllMods()
+					.stream()
+					.map(modContainer -> modContainer.getMetadata().getId())
+					.map(modid -> getModFileById(modid).getFile().getScanResult())
```
coderbot16, line 58: Won't this duplicate annotation data due to multiple mods sharing the same annotation data (jar-in-jar)? I think the root problem is that we associate one mod file with one mod ID, instead of the mod IDs within the same file sharing the same ModFileInfo like they would in Forge. We might need a way to indicate in general when a given JiJ mod is just a dummy mod for metadata purposes with something like `patchwork:parent` or `patchwork:dummy` on the patcher side.

qouteall: Changed to make jij mod and parent mod share the same ModFileInfo

coderbot16:
I think that it would still be best to have a proper custom property for locating the "true" mod file for a generated "dummy" jij mod. Using the `modmenu:parent` property is kind of an abuse of the property's meaning.

patchwork-fml/src/main/java/net/minecraftforge/fml/ModList.java
```diff
@@ -33,4 +47,60 @@ public boolean isLoaded(String modId) {
...
+		return allScanDataCache;
+	}
+
+	//return null if it's a Fabric mod
+	public static String getAnnotationHolderModid(String modid) {
+		ModContainer modContainer = FabricLoader.getInstance().getModContainer(modid).orElseThrow(
+				() -> new RuntimeException("No Mod Container for " + modid)
+		);
+
+		if (!ModFileInfo.isForgeMod(modContainer)) {
```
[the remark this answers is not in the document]
qouteall: Fabric mod and parent forge mod both miss patchwork:parent so I think that it's needed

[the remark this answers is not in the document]
coderbot16: But patchwork mods without a parent will contain `patchwork:annotations`. Maybe the code should check for that.

patchwork-fml/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFileInfo.java
```diff
@@ -0,0 +1,52 @@
...
+	//create empty mod file info for Fabric mods
+	public ModFileInfo() {
+		modFile = new ModFile();
+	}
+
+	public ModFile getFile() {
+		return modFile;
+	}
+
+	public static boolean isForgeMod(ModContainer modContainer) {
```
TheGlitch76: This isn't a blocking complaint but I don't like how we're calling mods "Forge" or "Fabric" here because there's nothing stopping a fabric mod from using Patchwork's APIs.

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,239 @@
...
+								argMethod.invoke(annotationObject)
+						);
+					}
+				}
+			} catch (Throwable e) {
+				LOGGER.catching(e);
+			}
+		}
+
+		private static boolean isArgumentMethod(Method method) {
```
TheGlitch76, line 193: This name is confusing... what do you mean by an "argument method"?

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,239 @@
...
+			switch (targetType) {
+			case TYPE:
+				return clazzObj.getAnnotation(annotationType);
+			case FIELD:
+				return clazzObj.getField(memberName)
+						.getAnnotation(annotationType);
+			case METHOD:
+				//TODO handle overloaded methods
+				String methodName = memberName.substring(0, memberName.indexOf('('));
+				return Arrays.stream(clazzObj.getDeclaredMethods())
```
TheGlitch76, line 214: This TODO should either be solved or alternatively assert that the stream's iterator size is 1 after the filter.

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,239 @@
...
+						.filter(method -> method.getName().equals(methodName))
+						.findFirst()
+						.orElseThrow(() -> new RuntimeException("Cannot get method " + memberName))
+						.getAnnotation(annotationType);
+			default:
+				return null;
```
TheGlitch76, line 220: Throw an AssertionError here instead of potentially causing an NPE

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
```diff
@@ -0,0 +1,258 @@
...
+	private Set<AnnotationData> annotationData;
+
+	public ModFileScanData(String annotationHolderModid) {
+		this.modid = annotationHolderModid;
+	}
+
+	//create empty mod file scan data for Fabric mods
+	public ModFileScanData() {
+		initialized = true;
+		annotationData = new HashSet<>();
```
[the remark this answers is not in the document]
TheGlitch76: Collections' empty sets are immutable, could that cause problems?

patchwork-fml/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModFile.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ModFile {
+	private final ModFileScanData modFileScanData;
+
+	public ModFile(String modid) {
+		modFileScanData = new ModFileScanData(modid);
+	}
+
+	//create empty mod file for Fabric mods
+	public ModFile() {
```
TheGlitch76, line 34: same thing about discriminating by forge/fabric

patchwork-fml/src/main/java/net/minecraftforge/fml/ModList.java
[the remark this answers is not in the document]
qouteall: It handles that.
And if the forge check is removed then when using getAllScanData it will say missing annotation  for Fabric mods

patchwork-fml/src/main/java/net/minecraftforge/forgespi/language/ModFileScanData.java
qouteall: argument method is the annotation object's method to get the annotation argument

qouteall:
I will let you fix the merge conflict.

coderbot16:
Superseded by #65

PR #64, opened 2020-05-05
Feature/iforgeblock

Nuclearfarts:
Stub implementation of IForgeBlock. This should keep a _lot_ of mods from `NoSuchMethodError`ing. As of right now, there isn't an implementation of everything. Some methods are commented out due to dependency on other parts of Forge we haven't implemented yet. Other methods are marked with a TODO comment listing the patch and Forge API call locations that must be implemented.


patchwork-extensions-block/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
```diff
@@ -0,0 +1,1069 @@
...
+import net.minecraft.world.World;
+import net.minecraft.world.dimension.TheEndDimension;
+import net.minecraft.world.explosion.Explosion;
+
+import net.fabricmc.api.EnvType;
+import net.fabricmc.api.Environment;
+
+import net.patchworkmc.mixin.extensions.block.FireBlockAccessor;
+
+public interface IForgeBlock {
```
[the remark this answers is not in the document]
Nuclearfarts: I'm not really sure how to do that without, uh, manually checking every single block class, which I'm definitely not doing.

settings.gradle
```diff
@@ -30,3 +30,4 @@ include 'patchwork-networking'
 include 'patchwork-networking-messages'
 include 'patchwork-vanilla-patches'
 include 'patchwork-enum-hacks'
+include 'patchwork-extensions-block'
```
TheGlitch76, line 33: This is supposed to be alphabetical. Can you fix enum-hacks while you're at it?

patchwork-extensions-block/src/main/java/net/minecraftforge/common/extensions/IForgeBlock.java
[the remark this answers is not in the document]
TheGlitch76: It's not too difficult to hunt those down (Especially in IDEA) when doing the patches so it shouldn't be a show-stopper for this PR.

TheGlitch76:
Well I forgot to mark as approved so I guess I'll just wait until its fixed.

settings.gradle
Nuclearfarts: none of this seems particularly alphabetical at all, considering "gui" and "networking" come after "registries"

coderbot16: well it *should* be alphabetical, but yeah that should probably be fixed after the PR is merged

patchwork-extensions-block/src/main/resources/patchwork-extensions-block.mixins.json
```diff
@@ -6,7 +6,8 @@
     "MixinBlock",
     "MixinBlockState",
     "BlockTagsAccessor",
-    "FireBlockAccessor"
+    "FireBlockAccessor",
+    "MixinBedBlock"
```
coderbot16, line 10: alphabetical order please

PR #67, opened 2020-05-18
Better initialization errors

coderbot16:
Errors during entrypoint invocation will be added as suppressed errors, for example.


patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
```diff
@@ -76,19 +77,41 @@ public void onInitialize() {
...
 		// Construct forge mods
 
-		for (ForgeInitializer initializer : FabricLoader.getInstance().getEntrypoints("patchwork", ForgeInitializer.class)) {
-			LOGGER.info("Constructing Forge mod: " + initializer.getModId());
+		List<ForgeInitializer> entrypoints;
+
+		try {
+			entrypoints = FabricLoader.getInstance().getEntrypoints("patchwork", ForgeInitializer.class);
+		} catch (Throwable t) {
+			throw new RuntimeException("Failed to get Patchwork entrypoints!", t);
```
TheGlitch76, line 85: While you're at it either make a dedicated exception or use an existing one.

PR #71, opened 2020-05-20
DataGatherEvent + ExistingFileHelper

Rongmario:
- DataGatherEvent in existing module `patchwork-events-lifecycle`
- ExistingFileHelper in a new module `patchwork-model`
- DataGatherEvent#DataGeneratorConfig#runAll is no-op for now.
- Further decisions can be made, such as injecting into `net.minecraft.data.Main` for generating data with patchwork too?
- Cleaned up comments

- Test mod/mechanism not yet decided.


l-Luna:
an implementation may be useful for people wanting to port forge mods with patchwork, but isn't needed for running mods, and IMO is something for when mods can be patched and expected to work more often than not.

Rongmario:
I see, yeah that would be better.

What do you think about LambdaExceptionUtils? It's used in runAll, and to be fair, would a try catch be fine?

` generators.forEach(LamdbaExceptionUtils.rethrowConsumer(DataGenerator::run)); `

coderbot16:
Yeah, I think a try catch is probably fine there.

patchwork-data-generators/src/main/java/net/minecraftforge/fml/event/lifecycle/GatherDataEvent.java
```diff
@@ -0,0 +1,117 @@
...
+	public static class DataGeneratorConfig {
+		private final Set<String> mods;
+		private final Path path;
+		private final Collection<Path> inputs;
+		private final boolean server;
+		private final boolean client;
+		private final boolean dev;
+		private final boolean reports;
+		private final boolean validate;
+		private List<DataGenerator> generators = new ArrayList<>();
```
coderbot16, line 84: This field can be final too
```suggestion
		private final List<DataGenerator> generators = new ArrayList<>();
```

patchwork-data-generators/src/main/java/net/minecraftforge/fml/event/lifecycle/GatherDataEvent.java
```diff
@@ -0,0 +1,117 @@
...
+			if (shouldExecute) generators.add(generator);
+			return generator;
+		}
+
+		public void runAll() {
+			for (DataGenerator generator : generators) {
+				try {
+					generator.run();
+				} catch (IOException e) {
+					e.printStackTrace();
```
coderbot16, line 112: It appears that the Forge implementation uses a [weird trick](https://stackoverflow.com/questions/27644361/how-can-i-throw-checked-exceptions-from-inside-java-8-streams/27644392#27644392) to throw this IOException in an unchecked manner. The way that it's done is kind of ugly though, I wonder if there's a way around it:

```java
@SuppressWarnings ("unchecked")
private static <E extends Throwable> void throwAsUnchecked(Exception exception) throws E { throw (E)exception; }
```

Technically we can just add a `throws IOException` on here since mods shouldn't be calling this method directly.

patchwork-data-generators/src/main/java/net/minecraftforge/fml/event/lifecycle/GatherDataEvent.java
```diff
@@ -0,0 +1,117 @@
...
+			this.validate = validate;
+		}
+
+		public Set<String> getMods() {
+			return mods;
+		}
+
+		public DataGenerator makeGenerator(final Function<Path, Path> pathEnhancer, final boolean shouldExecute) {
+			final DataGenerator generator = new DataGenerator(pathEnhancer.apply(path), inputs);
+			if (shouldExecute) generators.add(generator);
```
coderbot16, line 103:
```suggestion

			if (shouldExecute) {
				generators.add(generator);
			}

```

patchwork-data-generators/build.gradle
```diff
@@ -0,0 +1,7 @@
+archivesBaseName = "patchwork-data-generators"
+version = getSubprojectVersion(project, "0.1.0")
+
+dependencies {
+	compile project(path: ':patchwork-events-lifecycle', configuration: 'dev')
```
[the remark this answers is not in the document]
coderbot16: Ah, just wanted to make sure that it wasn't an unused import.

PR #72, opened 2020-05-23
Several simple helper methods

qouteall:


patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -0,0 +1,53 @@
...
+public class GameData {
+	/**
+	 * Check a name for a domain prefix, and if not present infer it from the
+	 * current active mod container.
+	 *
+	 * @param name          The name or resource location
+	 * @param warnOverrides If true, logs a warning if domain differs from that of
+	 *                      the currently currently active mod container
+	 *
+	 * @return The {@link net.minecraft.util.Identifier} with given or inferred domain
```
coderbot16, line 38: No need to fully qualify, it's already imported
```suggestion
	 * @return The {@link Identifier} with given or inferred domain
```

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -0,0 +1,53 @@
...
+import net.minecraft.util.Identifier;
+
+public class GameData {
+	/**
+	 * Check a name for a domain prefix, and if not present infer it from the
+	 * current active mod container.
+	 *
+	 * @param name          The name or resource location
+	 * @param warnOverrides If true, logs a warning if domain differs from that of
+	 *                      the currently currently active mod container
```
coderbot16, line 36: I know this is the forge documentation, but it doesn't paint the whole picture. If false, the prefix is forcefully updated without a warning, and if true, the prefix is not updated and there is just a warning message.

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -0,0 +1,53 @@
...
+		String oldPrefix = index == -1 ? "" : name.substring(0, index).toLowerCase(Locale.ROOT);
+		name = index == -1 ? name : name.substring(index + 1);
+		String prefix = ModLoadingContext.get().getActiveNamespace();
+
+		if (warnOverrides && !oldPrefix.equals(prefix) && oldPrefix.length() > 0) {
+			LogManager.getLogger().info("Potentially Dangerous alternative prefix `{}` for name `{}`, expected `{}`. This could be a intended override, but in most cases indicates a broken mod.", oldPrefix, name, prefix);
+			prefix = oldPrefix;
+		}
+
+		return new Identifier(prefix, name);
```
coderbot16, line 51: There's a version of this code in `ExtendedForgeRegistryEntry` already, it would be appreciated if you could deduplicate it.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistryEntry.java
```diff
@@ -31,6 +31,14 @@ public final IForgeRegistryEntry setRegistryName(Identifier name) {
 		return this;
 	}
 
+	public final IForgeRegistryEntry setRegistryName(String name) {
+		return setRegistryName(new Identifier(name));
+	}
+
+	public final IForgeRegistryEntry setRegistryName(String modID, String name) {
+		return setRegistryName(modID + ":" + name);
```
coderbot16, line 39: This should just construct an `Identifier` directly, instead of combining the strings together only for them to be split again in the constructor of `Identifier`.

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
qouteall: You mean make ForgeRegistryEntry extend ExtendedForgeRegistryEntry?

coderbot16: That would be good too, but in this case it would be nice to move the logic to a helper class outside of GameData and ExtendedForgeRegistryEntry.

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -0,0 +1,53 @@
...
+	 * @param name          The name or resource location
+	 * @param warnOverrides If true, logs a warning if domain differs from that of
+	 *                      the currently currently active mod container
+	 *
+	 * @return The {@link net.minecraft.util.Identifier} with given or inferred domain
+	 */
+	public static Identifier checkPrefix(String name, boolean warnOverrides) {
+		int index = name.lastIndexOf(':');
+		String oldPrefix = index == -1 ? "" : name.substring(0, index).toLowerCase(Locale.ROOT);
+		name = index == -1 ? name : name.substring(index + 1);
```
TheGlitch76, line 43: If this is the version of the impl that is kept, instead of spamming ternaries remove them and replace with an if( index == -1) or similar. Should make the code more readable.

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
qouteall: I found that making ForgeRegistryEntry extend ExtendedForgeRegistryEntry is wrong because it turns class method into interface method. And GameData is already a helper class. ExtendedForgeRegistryEntry.setRegistryName unlike this one does not replace the namespace so these two cannot be generalized.

qouteall: oh it can be generalized but it will make the code weird

PR #76, opened 2020-06-06
Fix annotation scanning argument object types

qouteall:


TheGlitch76:
looks like git history is getting messed up by our squash merge policy.
It's not a blocker here, but in the future please work on separate branches so that actual added commits are clear.

patchwork-fml/src/main/java/net/minecraftforge/fml/loading/moddiscovery/ModAnnotation.java
```diff
@@ -0,0 +1,133 @@
...
+		return MoreObjects.toStringHelper("Annotation")
+				.add("type", this.type).add("name", this.asmType.getClassName())
+				.add("member", this.member).add("values", this.values).toString();
+	}
+
+	public ElementType getType() {
+		return this.type;
+	}
+
+	public Type getASMType() {
```
coderbot16, line 76: The methods that are just used for the ASM annotation scanning code can be pruned, since they're probably not needed otherwise.

PR #79, opened 2020-06-13
Add custom registry support

rikka0w0:
Add custom registry support
Add net.minecraftforge.event.RegistryEvent$NewRegistry event


patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
```diff
@@ -63,7 +64,10 @@ private static void dispatch(Map<ForgeInitializer, FMLModContainer> mods, Functi
 		for (FMLModContainer container : mods.values()) {
 			ModLoadingContext.get().setActiveContainer(container, new FMLJavaModLoadingContext(container));
 
-			container.getEventBus().post(provider.apply(container));
+			Event event = provider.apply(container);
+			LOGGER.info("Firing event for modid " + container.getModId() + " : " + event.toString());
```
TheGlitch76, line 68: Use placeholder `{}` instead of string concatenation

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -109,42 +102,49 @@
...
+		wrap("dimension_type", DimensionType.class);
+		wrap("custom_stat", Identifier.class);
 		wrap("recipe_type", RecipeType.class);
 		wrap("rule_test", RuleTestType.class);
-		SCHEDULES = wrap("schedule", Schedule.class);
-		SENSOR_TYPES = wrap("sensor_type", SensorType.class);
-		SOUND_EVENTS = wrap("sound_event", SoundEvent.class);
-		STAT_TYPES = wrap("stat_type", StatType.class);
-		wrap("structure_feature", StructureFeature.class);
+		/*wrap("structure_feature", StructureFeature.class);
```
TheGlitch76, line 144: Why comment this out when it wasn't previously?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -19,189 +19,804 @@
...
+	private int getIDRaw(Identifier name) {
+		return getIDRaw(this.names.get(name));
+	}
+
+	public V getValue(int id) {
+		V ret = this.ids.get(id);
+		return ret == null ? this.defaultValue : ret;
+	}
+
+	void validateKey() {
```
TheGlitch76, line 260: What's up with these package-private members? If they're unused they can safely be removed.
It would also reduce the noise of this PR a fair bit.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -19,189 +19,804 @@
...
+
+		while (ret == null && key != null) {
+			ret = this.names.get(key);
+			key = this.aliases.get(key);
+		}
+
+		return ret;
+	}
+
+	@Deprecated // Public for ByteByfUtils only!
```
TheGlitch76, line 385: remove if unused, change the comment to a javadoc otherwise

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
rikka0w0: These types don't have a registry in Forge.
In particular, StructureFeature extends Feature, so StructureFeature cannot be a registry base class, as this violates forge's registry rule: a new registry type should not be a sub class of an existing registry type

Please correct me if these are not true.

patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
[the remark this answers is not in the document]
rikka0w0: Yes, sure. 
Btw, how to set the fabric / patchwork logging level to debug?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -19,189 +19,804 @@
...
+import net.minecraftforge.fml.ModLoadingContext;
+import net.minecraftforge.fml.common.ObfuscationReflectionHelper;
 
 import net.minecraft.util.Identifier;
-import net.minecraft.util.registry.DefaultedRegistry;
-import net.minecraft.util.registry.Registry;
 
-public class ForgeRegistry<V extends IForgeRegistryEntry<V>> implements IForgeRegistry<V> {
+// TODO: snapshot and missing mapping are not implemented
+public class ForgeRegistry<V extends IForgeRegistryEntry<V>> implements IForgeRegistryInternal<V>, IForgeRegistryModifiable<V> {
```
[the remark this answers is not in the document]
rikka0w0: Yes, I'm reverting back to the  vanilla registry system

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -153,16 +153,30 @@
 		Identifier identifier = new Identifier("minecraft", name);
 		Registry registry = Registry.REGISTRIES.get(identifier);
 
-		ForgeRegistry wrapped = new ForgeRegistry(identifier, registry, superClazz);
```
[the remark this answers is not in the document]
rikka0w0: In the latest commit, the entire ForgeRegistry impl is based on the vanilla registry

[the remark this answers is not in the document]
rikka0w0: ForgeRegistry#freeze, unfreeze, isLock are used for access control, prevent forge mods register objects at incorrect stage.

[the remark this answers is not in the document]
rikka0w0: ForgeRegistry#getSlaveMap and ForgeRegistry#setSlaveMap are for handing special vanilla case:
StructrueFeature extends Feature, this is not allowed in Forge, so they created "SlaveMap". Not sure if we should keep this.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -204,4 +272,72 @@ public int hashCode() {
...
+		if (!this.isVanilla) {
+			throw new UnsupportedOperationException("Vanilla registry does not support remove().");
+		}
+
+		throw new UnsupportedOperationException("Remove() is not implemented for Forge Mod registries.");
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public <T> T getSlaveMap(Identifier name, Class<T> type) {
```
rikka0w0, line 334: ForgeRegistry#getSlaveMap and ForgeRegistry#setSlaveMap are for handing special vanilla case:
StructrueFeature extends Feature, this is not allowed in Forge, so they created "SlaveMap". Not sure if we should keep this.

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -43,28 +53,121 @@ public String getName() {
...
-	 * @param registry the registry to add to the mapping
-	 */
-	void addRegistry(Identifier key, ForgeRegistry registry) {
-		registries.put(key, registry);
+	public Identifier getName(Class<?> clazz) {
+		if (clazz == null) {
+			return null;
+		}
+
+		do {
```
TheGlitch76, line 89: Ew.
Would changing this to a normal while loop and removing the extra null check above break any behavior in this method?

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -43,28 +53,121 @@ public String getName() {
...
+
+		return null;
+	}
+
+	public Set<Identifier> getRegistryNames() {
+		return Registry.REGISTRIES.getIds();
+	}
+
+	<V extends IForgeRegistryEntry<V>> ForgeRegistry<V> createRegistry(Identifier name, RegistryBuilder<V> builder) {
+		Set<Class<?>> parents = Sets.newHashSet();
```
TheGlitch76, line 107: This is a tiny nitpick, but please replace Sets.newHashSet() in this file with the diamond operator `new HashSet<>()`

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -40,17 +41,68 @@
...
 	public static Marker REGISTRIES = MarkerManager.getMarker("REGISTRIES");
 	private static Logger LOGGER = LogManager.getLogger();
-	private Identifier name;
-	private Registry<V> vanilla;
-	private Class<V> superType;
+	private final Identifier name;
+	private final boolean isVanilla;
+	private final Registry<V> vanilla;
+	private final Class<V> superType;
+	private final Map<Identifier, ?> slaves = Maps.newHashMap();
```
TheGlitch76, line 53: Diamond operator here please

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -40,17 +41,68 @@
...
+	 * @param vanilla
+	 * @param superType
+	 */
 	public ForgeRegistry(Identifier name, Registry<V> vanilla, Class<V> superType) {
 		this.name = name;
 		this.vanilla = vanilla;
 		this.superType = superType;
+
+		this.create = null;
+		this.add = null;
```
TheGlitch76, line 97: What would these callbacks be used for?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -40,17 +41,68 @@
...
+		this.clear = null;
+		this.stage = null;
+		this.isModifiable = false;
+		this.isVanilla = true;
+
+		if (this.create != null) {
```
TheGlitch76, line 103: `this.create != null` will always be true

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
TheGlitch76: Is it something that could be redirected in Patcher, or is it too much of a hack?

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
coderbot16: Object.class.getSuperclass() returns null, so that part stays the same. But, the existingKey code would be missed. Is there any reason that the existingKey stuff isn't the first thing in the method?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: These callbacks are set in the RegistryBuilder, when the ForgeRegistry is created. We should keep this to maintain compatibility with Forge

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: I'm also looking at it

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -0,0 +1,317 @@
...
+	private Identifier registryName;
+	private Class<T> registryType;
+	private Identifier optionalDefaultKey;
+	private int minId = 0;
+	private int maxId = MAX_ID;
+	private List<AddCallback<T>> addCallback = Lists.newArrayList();
+	private List<ClearCallback<T>> clearCallback = Lists.newArrayList();
+	private List<CreateCallback<T>> createCallback = Lists.newArrayList();
+	private List<ValidateCallback<T>> validateCallback = Lists.newArrayList();
+	private List<BakeCallback<T>> bakeCallback = Lists.newArrayList();
```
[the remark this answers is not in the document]
rikka0w0: We cannot use event here, these are not global callba cks, they are implemented by the registry to monitor the modification of its content

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ExtendedVanillaRegistry.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package net.patchworkmc.impl.registries;
+
+import net.minecraftforge.registries.ForgeRegistry;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+import net.minecraft.util.registry.SimpleRegistry;
+
+// TODO: implement default registry
```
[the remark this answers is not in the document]
rikka0w0: Yes

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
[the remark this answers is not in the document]
coderbot16: Just so we're clear, I'm referring to [fabric events](https://github.com/FabricMC/fabric/tree/1.16/fabric-api-base/src/main/java/net/fabricmc/fabric/api/event) here, not Forge events. Fabric events are basically a list (well, it's implemented using an array) of callbacks, which is exactly what's going on here.

patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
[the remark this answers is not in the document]
coderbot16: Not sure, it's probably some log4j setting. Maybe it's better to just remove these messages?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
coderbot16: I think it's possible to work with this system while still using our vanilla wrapper implementation, I think we can avoid patcher here.

patchwork-registries/src/main/java/net/patchworkmc/mixin/registries/MixinSimpleRegistry.java
```diff
@@ -0,0 +1,67 @@
...
+import net.minecraftforge.registries.ForgeRegistry;
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.SimpleRegistry;
+
+import net.patchworkmc.impl.registries.VanillaRegistry;
+
+@Mixin(SimpleRegistry.class)
+public abstract class MixinSimpleRegistry<T> implements VanillaRegistry {
```
coderbot16, line 36: Why not use `RegistryEntryAddedCallback` instead? It allows you to listen to registry additions quite easily. It would allow you to remove this mixin, VanillaRegistry, and ForgeRegistryProvider.

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ExtendedVanillaRegistry.java
```diff
@@ -0,0 +1,44 @@
...
+public class ExtendedVanillaRegistry<V extends IForgeRegistryEntry<V>> extends SimpleRegistry<V> implements ForgeRegistryProvider {
+	private final ForgeRegistry<V> forgeRegistry;
+	public ExtendedVanillaRegistry(ForgeRegistry<V> forgeRegistry) {
+		this.forgeRegistry = forgeRegistry;
+	}
+
+	public void clear() {
```
coderbot16, line 34: This should probably actually be implemented as a mixin to SimpleRegistry, and then have a ClearableRegistry interface implemented by the mixin.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -109,57 +105,57 @@
...
+		// TODO: Patchwork Project, check these
+		wrap("dimension_type", DimensionType.class);
+		wrap("custom_stat", Identifier.class);
 		wrap("recipe_type", RecipeType.class);
 		wrap("rule_test", RuleTestType.class);
-		SCHEDULES = wrap("schedule", Schedule.class);
-		SENSOR_TYPES = wrap("sensor_type", SensorType.class);
-		SOUND_EVENTS = wrap("sound_event", SoundEvent.class);
-		STAT_TYPES = wrap("stat_type", StatType.class);
 		wrap("structure_feature", StructureFeature.class);
```
coderbot16, line 147: Per the StructureFeature / Feature issue and similar issues with ForgeRegistry, will this crash or break mods as is? I would think that the overlap checks would cause this to fail or mess up somehow.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -204,4 +272,72 @@ public int hashCode() {
...
+	}
+
+	@Override
+	public void clear() {
+		if (!this.isModifiable) {
+			throw new UnsupportedOperationException("Attempted to clear a non-modifiable Forge Registry");
+		}
+
+		if (this.isLocked()) {
+			throw new IllegalStateException("Attempted to clear the registry to late.");
```
coderbot16, line 300: :bike: :house:
```suggestion
			throw new IllegalStateException("Attempted to clear the registry too late.");
```

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -204,4 +272,72 @@ public int hashCode() {
...
+		if (this.clear != null) {
+			this.clear.onClear(this, stage);
+		}
+
+		ExtendedVanillaRegistry<V> reg = (ExtendedVanillaRegistry<V>) this.vanilla;
+		reg.clear();
+	}
+
+	@Override
+	public V remove(Identifier key) {
```
coderbot16, line 316: Maybe have a TODO here?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -204,4 +272,72 @@ public int hashCode() {
...
+		if (!this.isModifiable) {
+			throw new UnsupportedOperationException("Attempted to clear a non-modifiable Forge Registry");
+		}
+
+		if (this.isLocked()) {
+			throw new IllegalStateException("Attempted to clear the registry to late.");
+		}
+
+		if (!this.isVanilla) {
+			throw new UnsupportedOperationException("Vanilla registry does not support clear().");
```
coderbot16, line 304: All registries should be able to support clearing with the help of a mixin (see my other comment)

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
coderbot16: You mean, it will always be false? This code looks like dead code to me.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -40,17 +41,68 @@
...
+		this.vanilla = new ExtendedVanillaRegistry<>(this);
+		this.isVanilla = false;
+
+		if (this.create != null) {
+			this.create.onCreate(this, stage);
+		}
+	}
+
+	/**
+	 * For vanilla registeries.
```
coderbot16, line 86:
```suggestion
	 * For vanilla registries.
```

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -40,17 +41,68 @@
...
+	private final CreateCallback<V> create;
+	private final AddCallback<V> add;
+	private final ClearCallback<V> clear;
+	private final RegistryManager stage;
+	private final boolean isModifiable;
+
+	private boolean isFrozen = false;
+
+	/**
+	 * Called by RegistryBuilder, for moded registries.
```
coderbot16, line 63:
```suggestion
	 * Called by RegistryBuilder, for modded registries.
```

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -40,17 +41,68 @@
...
+	public ForgeRegistry(RegistryManager stage, Identifier name, RegistryBuilder<V> builder) {
+		this.stage = stage;
+		this.name = name;
+		this.superType = builder.getType();
+		this.create = builder.getCreate();
+		this.add = builder.getAdd();
+		this.clear = builder.getClear();
+		this.isModifiable = builder.getAllowModifications();
 
+		this.vanilla = new ExtendedVanillaRegistry<>(this);
```
coderbot16, line 77: Maybe use a defaulted registry here if the builder declares default values / keys?

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -43,28 +53,121 @@ public String getName() {
...
-		return (Class<V>) RegistryClassMapping.getClass(key);
+		return (Class<V>) this.superTypes.inverse().get(key);
 	}
 
 	@SuppressWarnings("unchecked")
 	public <V extends IForgeRegistryEntry<V>> ForgeRegistry<V> getRegistry(Identifier key) {
-		return (ForgeRegistry<V>) this.registries.get(key);
+		Registry<?> vanillaRegistry = Registry.REGISTRIES.get(key);
+
+		if (vanillaRegistry instanceof ForgeRegistryProvider) {
```
coderbot16, line 63: Alternatively, it might be a good idea to have a vanilla registry instance -> ForgeRegistry map here, instead of attaching it as a value of the registry instance with a mixin? Not entirely sure on this issue though.

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
coderbot16: Okay, after looking closely, it would only change the behavior if there's a registry with a super type of Object. That shouldn't happen, and in fact, we should probably have checks for that in the creation methods.

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -43,28 +53,121 @@ public String getName() {
...
+
+	<V extends IForgeRegistryEntry<V>> ForgeRegistry<V> createRegistry(Identifier name, RegistryBuilder<V> builder) {
+		Set<Class<?>> parents = Sets.newHashSet();
+		findSuperTypes(builder.getType(), parents);
+		SetView<Class<?>> overlappedTypes = Sets.intersection(parents, superTypes.keySet());
+
+		if (!overlappedTypes.isEmpty()) {
+			Class<?> foundType = overlappedTypes.iterator().next();
+			throwSuperClassOverlapException(foundType, builder.getType());
+		}
```
coderbot16, line 114: Since most of this code is shared with wrapVanilla, could it be moved into a common method?

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -43,28 +53,121 @@ public String getName() {
...
+		Registry<V> registry = (Registry<V>) Registry.REGISTRIES.get(identifier);
+
+		Set<Class<?>> parents = Sets.newHashSet();
+		findSuperTypes(superClazz, parents);
+		SetView<Class<?>> overlappedTypes = Sets.intersection(parents, superTypes.keySet());
+
+		if (!overlappedTypes.isEmpty()) {
+			Class<?> foundType = overlappedTypes.iterator().next();
+
+			if (overlappedTypes.size() == 1) {
```
coderbot16, line 157: Should explain the significance of this check. What's going on here? Also, how does this handle stuff like Feature/StructureFeature?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
TheGlitch76: Yes, that was a typo

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
TheGlitch76: This should be fired then.
I believe coderbot's comment on RegistryEntryAddedCallback would resolve this.

coderbot16:
Also, almost forgot, but can we get a minor version bump on the module? Thanks!

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
rikka0w0: If I understand correctly, forge mods register StructureFeatures during RegistryEvent.Register<Feature> and forge will place them into the correct vanilla registry. So yes, we need to do more stuff on this. I'm thinking about implementing forge's slavemap. The slavemap only affects forge mods, fabric mods are not affected

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ExtendedVanillaRegistry.java
rikka0w0: This class is for forge modded registries, they may need functions that is not supported by SimpleRegistry, like remove(). I'm planning to reimpl this

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: Could clearing vanilla registries break something?

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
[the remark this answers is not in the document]
rikka0w0: addCallback is now implemented with the aid of fabric events

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
rikka0w0: In the vanilla registry record, all forge modded registries are instance of ExtendedVanillaRegistry (super class of MutableRegistry), and they hold the reference to the its ForgeRegistry. The interface ForgeRegistryProvider is used for both forge and vanilla registries for retrieving ForgeRegistry reference.

Having a vanilla registry->ForgeRegistry map would work, but I still prefer to not maintaining a separate map.

rikka0w0: I also have a concern, currently vanilla and fabric mods can access registries added by forge, but not the other way around. How should we handle the registries created by fabric mods? Should we make them visible to forge mods?

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
rikka0w0: wrapVanilla has been removed

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: Forge mods register both Feature and StructureFeature during IForgeRegistry<Feature<?>> registry event:
https://github.com/Cheaterpaul/Vampirism/blob/1.14/src/main/java/de/teamlapen/vampirism/core/ModFeatures.java

patchwork-registries/src/main/java/net/patchworkmc/mixin/registries/MixinSimpleRegistry.java
rikka0w0: RegistryEntryAddedCallback is used in the latest commit, the unnecessary mixin injection is removed

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/RegistryEventDispatcher.java
```diff
@@ -26,36 +26,26 @@
...
 	 */
 	private static List<Identifier> getExpectedOrdering() {
-		List<Identifier> registries = new ArrayList<>(Registry.REGISTRIES.getIds());
+		List<Identifier> registries = new ArrayList<>(RegistryManager.ACTIVE.getRegistryNames());
+		registries.sort((o1, o2) -> String.valueOf(o1).compareToIgnoreCase(String.valueOf(o2)));
 
 		registries.remove(Registry.REGISTRIES.getId(Registry.BLOCK));
 		registries.remove(Registry.REGISTRIES.getId(Registry.ITEM));
 
-		registries.sort((o1, o2) -> String.valueOf(o1).compareToIgnoreCase(String.valueOf(o2)));
```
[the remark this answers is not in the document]
rikka0w0: It has been reverted

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: getSlaveMap is declared in IForgeRegistry, so it is a part of the Forge API, we should keep this to maintain compatibility

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
rikka0w0: I have changed the loop to a standard while loop

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: In the latest commit, the mixin injection is replaced by a RegistryEntryAddedCallback

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: this.create can be null, if RegistryBuilder does not supply any callbacks. Upon registry creation, RegistryBuilder batch all callbacks into one. So here we need to check the nullity.

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ExtendedVanillaRegistry.java
rikka0w0: ClearableRegistry is created

patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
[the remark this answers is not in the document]
rikka0w0: LOGGER.info are changed to LOGGER.debug

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
rikka0w0: vanilla wrapping is done by the RegistryManager and GameData

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ExtendedVanillaRegistry.java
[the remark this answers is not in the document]
rikka0w0: But for this minor version, let's just use the functions provided by SimpleRegistry. They should be able to covert most Forge mod usage. After this pr, I'm going to reimpl this class to give full forge compatibility

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
rikka0w0: fixed

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -39,18 +40,84 @@
...
+			this.isVanilla = false;
+		} else {
+			// Vanilla registry
+			this.vanilla = vanilla;
+			((VanillaRegistry) this.vanilla).setForgeRegistry(this);
+			this.isVanilla = true;
+		}
+
+		// Fabric hooks
+		// TODO: Some vanilla registry types are not patched yet, add this check to avoid crash
```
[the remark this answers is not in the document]
rikka0w0: DimentionTypes

and some vanilla registry should not be patched, e.g. custom_stat is Identifier

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -39,18 +40,84 @@
...
+
+		if (this.create != null) {
+			this.create.onCreate(this, stage);
+		}
+	}
+
+	@Override
+	public void onEntryAdded(int rawId, Identifier id, V newValue) {
+		if (this.isLocked()) {
+			throw new IllegalStateException(String.format("The object %s (name %s) is being added too late.", newValue, id));
```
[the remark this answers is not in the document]
rikka0w0: Sure, I will add a note

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
rikka0w0: Perhaps we should leave them there for now, when we start working on the ModDimention, we can have a look at DimentionType registry together

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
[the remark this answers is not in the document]
rikka0w0: This class is just a builder for constructing ForgeRegistry.
During the creation, references to callback handlers are added by forge mods and forge itself.
Upon the creation of ForgeRegistry, it combines the list into one single xxxCallback handler and the resulting handler is stored inside ForgeRegistry.

E.g.:
https://github.com/rikka0w0/patchwork-api/blob/22f059b0959ec6b9079af40d4a5496f55aa0bdef/patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java#L213

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -0,0 +1,334 @@
...
+		return this.vanilla;
+	}
+
+	public RegistryBuilder<T> setName(Identifier name) {
+		this.registryName = name;
+		return this;
+	}
+
+	public RegistryBuilder<T> setType(Class<T> type) {
+		this.registryType = type;
```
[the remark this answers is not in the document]
rikka0w0: the RegistryManger#findSuperTypes cannot block Object.class and null.
This must be a forge's problem. We can do a check here, if the type is not valid, we can throw an error.

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -43,28 +52,85 @@ public String getName() {
...
+					"Found existing registry of type {} named {}, you cannot create a new registry ({}) with type {}, as {} has a parent of that type",
+					foundType, superTypes.get(foundType), name, builder.getType(), builder.getType());
+			throw new IllegalArgumentException(
+					"Duplicate registry parent type found - you can only have one registry for a particular super type");
+		}
+
+		ForgeRegistry<V> reg = new ForgeRegistry<V>(this, name, builder);
+		superTypes.put(builder.getType(), name);
+
+		return getRegistry(name);
```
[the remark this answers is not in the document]
rikka0w0: This were left from previous implementation, removed

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -0,0 +1,334 @@
...
+
+	public RegistryBuilder<T> disableSaving() {
+		this.saveToDisc = false;
+		return this;
+	}
+
+	public RegistryBuilder<T> disableSync() {
+		this.sync = false;
+		return this;
+	}
```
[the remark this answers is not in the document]
rikka0w0: Added warning message

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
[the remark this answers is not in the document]
rikka0w0: an error will be thrown if type is null or Object

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -39,18 +40,84 @@
...
+	private V oldValue; // context of AddCallback, does not use elsewhere
+
+	/**
+	 * Called by RegistryBuilder, for modded registries.
+	 * @param stage
+	 * @param name
+	 * @param builder
+	 */
+	@SuppressWarnings("unchecked")
+	public ForgeRegistry(RegistryManager stage, Identifier name, RegistryBuilder<V> builder) {
```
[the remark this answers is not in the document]
rikka0w0: changed to protected

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
```diff
@@ -150,16 +146,7 @@
...
-		Registry registry = Registry.REGISTRIES.get(identifier);
-
-		ForgeRegistry wrapped = new ForgeRegistry(identifier, registry, superClazz);
-
-		RegistryClassMapping.register(wrapped);
-		RegistryManager.ACTIVE.addRegistry(identifier, wrapped);
-		RegistryEventDispatcher.register(wrapped);
-
-		return wrapped;
+		return RegistryManager.ACTIVE.getRegistry(superClazz);
```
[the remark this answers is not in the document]
rikka0w0: I can clean up this part later

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -70,17 +137,21 @@ public void register(V value) {
...
-						LOGGER.warn(REGISTRIES, "Registry {}: The object {} has been registered twice for the same name {}.", this.superType.getSimpleName(), value, identifier);
-
-						return;
-					} else {
-						throw new IllegalArgumentException(String.format("The name %s has been registered twice, for %s and %s.", identifier, oldValue, value));
-					}
-				}
-		);
+		if (potentialOldValue.isPresent()) {
+			V oldValue = potentialOldValue.get();
```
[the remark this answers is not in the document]
rikka0w0: I need to do multiple checks here, I feel it is easier to control the flow when ifPresent() is removed here.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -70,17 +137,21 @@ public void register(V value) {
...
+
+			if (oldValue == value) {
+				LOGGER.warn(REGISTRIES, "Registry {}: The object {} has been registered twice for the same name {}.", this.superType.getSimpleName(), value, identifier);
+				return;
+			} else if (this.allowOverrides) {
+				this.oldValue = oldValue;
+				LOGGER.debug(REGISTRIES, "Registry {}: The object {} {} has been overridden by {}.", this.superType.getSimpleName(), identifier, oldValue, value);
```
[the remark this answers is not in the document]
rikka0w0: By default, allowOverrides is set to false by the RegistryBuilder. All wrapped vanilla registeries have allowOverrides set to false, so this wont be a problem.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -204,4 +282,61 @@ public int hashCode() {
...
+		if (!this.isModifiable) {
+			throw new UnsupportedOperationException("Attempted to clear a non-modifiable Forge Registry");
+		}
+
+		if (this.isLocked()) {
+			throw new IllegalStateException("Attempted to clear the registry too late.");
+		}
+
+		if (this.isVanilla) {
+			LOGGER.debug("Vanilla registery {} is cleared!", this.name);
```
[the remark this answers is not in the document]
rikka0w0: I think we should keep this logic, and throw a warning message when the vanilla registry is replaced or cleared

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
rikka0w0: I cant imagin any of those cases... So our current code should be fine

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -58,4 +213,17 @@ public static Identifier checkPrefix(String name, boolean warnOverrides) {
...
 		return new Identifier(prefix, newName);
 	}
+
+	@SuppressWarnings("unchecked")
+	public static <V extends IForgeRegistryEntry<V>> IForgeRegistry<V> wrapVanilla(Identifier identifier, Registry<?> registry) {
+		RegistryBuilder<V> builder = (RegistryBuilder<V>) GameData.getBuilder(identifier);
+
+		if (builder.getType().isAssignableFrom(StructureFeature.class)) {
+			// TODO: StructureFeature Needs slave map impl
+			return null;
```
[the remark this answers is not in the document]
rikka0w0: I'm planning to do so

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -19,14 +19,169 @@
...
+	// public static final Identifier SERIALIZERS = new
+	// Identifier("minecraft:dataserializers");
+	// public static final Identifier LOOT_MODIFIER_SERIALIZERS = new
+	// Identifier("forge:loot_modifier_serializers");
+
+	private static final Map<Identifier, RegistryBuilder<?>> vanillaWrapperBuilders = new HashMap<>();
+
+	static {
+		if (!vanillaWrapperBuilders.isEmpty()) {
+			throw new RuntimeException("Duplicated type map initialization!");
```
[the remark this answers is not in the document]
rikka0w0: just in case(

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -19,14 +19,169 @@
...
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	private static RegistryBuilder<?> wrap(Identifier id, Class superClazz) {
+		RegistryBuilder builder = new RegistryBuilder();
+		builder.setName(id);
+		builder.setType(superClazz);
+		vanillaWrapperBuilders.put(id, builder);
+		return builder;
+	}
+
+	public static RegistryBuilder<?> getBuilder(Identifier identifier) {
```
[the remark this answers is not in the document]
rikka0w0: changed to private

patchwork-registries/src/main/java/net/patchworkmc/mixin/registries/MixinRegistry.java
```diff
@@ -0,0 +1,41 @@
...
+import net.minecraftforge.registries.GameData;
+
+import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.MutableRegistry;
+import net.minecraft.util.registry.Registry;
+
+@Mixin(Registry.class)
+public class MixinRegistry {
+	@Inject(at = @At("TAIL"), method = "putDefaultEntry(Ljava/lang/String;Lnet/minecraft/util/registry/MutableRegistry;Ljava/util/function/Supplier;)Lnet/minecraft/util/registry/MutableRegistry;")
+	private static <T, R extends MutableRegistry<T>> void putDefaultEntry(String name, R reg, Supplier<T> supplier, CallbackInfoReturnable<R> info) {
```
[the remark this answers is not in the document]
rikka0w0: changed

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -58,4 +213,17 @@ public static Identifier checkPrefix(String name, boolean warnOverrides) {
...
+	@SuppressWarnings("unchecked")
+	public static <V extends IForgeRegistryEntry<V>> IForgeRegistry<V> wrapVanilla(Identifier identifier, Registry<?> registry) {
+		RegistryBuilder<V> builder = (RegistryBuilder<V>) GameData.getBuilder(identifier);
+
+		if (builder.getType().isAssignableFrom(StructureFeature.class)) {
+			// TODO: StructureFeature Needs slave map impl
+			return null;
+		}
+
+		builder.setVanillaRegistry((Registry<V>) registry);
```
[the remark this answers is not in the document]
rikka0w0: yes, this can happen.
I add a check in the latest commit so it wont crash.
However we should think about making Fabric registry visible to forge mods

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -19,14 +19,169 @@
...
+
+	private static void wrap(String name, Class superClazz) {
+		wrap(new Identifier(name), superClazz);
+	}
+
+	@SuppressWarnings({"unchecked", "rawtypes"})
+	private static RegistryBuilder<?> wrap(Identifier id, Class superClazz) {
+		RegistryBuilder builder = new RegistryBuilder();
+		builder.setName(id);
+		builder.setType(superClazz);
```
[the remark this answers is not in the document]
rikka0w0: override is not enabled by default, so this is fine

patchwork-registries/src/main/java/net/minecraftforge/registries/GameData.java
```diff
@@ -19,14 +19,169 @@
...
+import net.minecraft.world.gen.decorator.Decorator;
+import net.minecraft.world.gen.feature.Feature;
+import net.minecraft.world.gen.feature.StructureFeature;
+import net.minecraft.world.gen.surfacebuilder.SurfaceBuilder;
+import net.minecraft.world.poi.PointOfInterestType;
 
 public class GameData {
+	// These are needed because some Forge mods access these static public fields
+	// Vanilla registries
+	// Names used here match those in net.minecraft.util.Registry
```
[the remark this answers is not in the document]
rikka0w0: understood, removed

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/VanillaRegistry.java
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.patchworkmc.impl.registries;
+
+import net.minecraftforge.registries.ForgeRegistry;
+
+public interface VanillaRegistry extends ForgeRegistryProvider, ClearableRegistry {
```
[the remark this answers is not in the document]
rikka0w0: We can leave it for now and throw a warning when a forge mod perform this action

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
rikka0w0: note added

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
rikka0w0: We can leave DimentionType for now, adding support for modded dimention is beyond this pr.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistries.java
rikka0w0: StructureFeatures registration for forge mods is now supported

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: StructureFeature is supported

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: defaulted registry is supported now

rikka0w0:
I know this pr is getting longer, but we do need to have this much code to make the forge registry functional.
Here are features implemented:
1. NewRegistry event
1. Default registry entry
1. Minimal patch to the vanilla registry system, use vanilla registry system to store forge mod registries, so they can be accessed by fabric mods
1. Forge getSlaveMap() support, for compatibility

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -0,0 +1,347 @@
...
+	private int minId = 0;
+	private int maxId = MAX_ID;
+	private List<AddCallback<T>> addCallback = Lists.newArrayList();
+	private List<ClearCallback<T>> clearCallback = Lists.newArrayList();
+	private List<CreateCallback<T>> createCallback = Lists.newArrayList();
+	private List<ValidateCallback<T>> validateCallback = Lists.newArrayList();
+	private List<BakeCallback<T>> bakeCallback = Lists.newArrayList();
+	private boolean saveToDisc = true;
+	private boolean sync = true;
+	private boolean allowOverrides = true;
```
coderbot16, line 56: > override is not enabled by default, so this is fine
It appears that it is enabled by default. It needs to be disabled on vanilla registries for now.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
coderbot16: At this point it might make more sense to just do `this.oldValue = vanilla.getOrEmpty(identifier).orElse(null);` and then work with the nullable value directly, since that's effectively what's going on here. It would probably simplify the control flow a bit.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
coderbot16: This is not the case, see my comment on RegistryBuilder.

patchwork-registries/README.md
```diff
@@ -0,0 +1,25 @@
+# Implemented features
+1. `RegistryEvent.NewRegistry` for allowing forge mods to add their own registry
+1. `RegistryEvent.Register` for different in-game object registration and forge mods' custom registry
```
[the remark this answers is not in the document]
rikka0w0: It is actually a markdown files, it looks better on a proper viewer or on github

patchwork-registries/README.md
```diff
@@ -0,0 +1,25 @@
...
+1. StructureFeature registration needs more testing
+1. A fully modifiable Forge mod registry implementation
+1. Code clean up in ForgeRegistries, the entry order should match Forge's version.
+1. Custom forge registries and Patchwork's wrapper implementation, see GameData:
+```
+public static final Identifier MODDIMENSIONS = new Identifier("forge:moddimensions");
+public static final Identifier SERIALIZERS = new Identifier("minecraft:dataserializers");
+public static final Identifier LOOT_MODIFIER_SERIALIZERS = new Identifier("forge:loot_modifier_serializers");
+```
+1. Make fabric registries visible to Forge mods (Carefully consider this!)
```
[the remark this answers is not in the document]
rikka0w0: I got the point, so ForgeRegistry will only wrap vanilla registries and handle forge mod registries, forge mods cannot access fabric registry via ForgeRegistry, but they can access them via the vanilla registry system. This will make our life a lot easier

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
rikka0w0: I see, gonna to fix that in the next commit

patchwork-registries/README.md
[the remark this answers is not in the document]
coderbot16: I understand that, but I like to keep my markdown files readable in plain text editors too.

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ForgeModRegistryImpl.java
```diff
@@ -0,0 +1,122 @@
...
+import net.minecraft.util.registry.Registry;
+
+import net.fabricmc.fabric.api.event.registry.RegistryEntryAddedCallback;
+
+/**
+ * This interface is for avoiding duplicated implementation as much as possible
+ * The vanilla field {@link net.minecraft.util.registry.SimpleRegistry#indexedEntries} is not used.
+ * @author Rikka0w0
+ */
+public interface ForgeModRegistryImpl<V extends IForgeRegistryEntry<V>> extends ForgeRegistryProvider, ModifiableRegistry<V> {
```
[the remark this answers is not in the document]
rikka0w0: The "entries" is a BiMap<int, Object> implemented by vanilla, it is used in the registry system and other places. It supports put(named set in the code) and clear, but it does not support remove. There is an alternative implement, use mixin to add a remove() method for that class. Or we can implement our own bimap which extends the vanilla BiHashMap, and use mixin to replace the instance of "entries". 
I do agree that the bitset can be replaced by a nextId and a list of available Ids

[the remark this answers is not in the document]
coderbot16: In that case, why not replace it in the constructor with an instance of a class extending Int2ObjectBiMap that does support removal (As you've said)?

> Or we can implement our own bimap which extends the vanilla BiHashMap, and use mixin to replace the instance of "entries

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
rikka0w0: fixed in the latest commit

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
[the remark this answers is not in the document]
rikka0w0: solved

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ModifiableRegistry.java
```diff
@@ -19,6 +19,9 @@
 
 package net.patchworkmc.impl.registries;
 
-public interface ClearableRegistry {
+import net.minecraft.util.Identifier;
+
+public interface ModifiableRegistry<V> {
```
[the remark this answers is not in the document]
rikka0w0: rename to PatchworkModifiableRegistry?

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -310,22 +343,32 @@ public void clear() {
...
+		if (!this.isModifiable) {
+			throw new UnsupportedOperationException("Attempted to remove from a non-modifiable Forge Registry");
+		}
+
+		if (this.isLocked()) {
+			throw new IllegalStateException("Attempted to remove from the registry too late.");
+		}
+
+		// If it is modifiable, it must be a forge mod registry, vanilla registries do not support remove().
+		V removed = ((ModifiableRegistry<V>) this.vanilla).remove(key);
```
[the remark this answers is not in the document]
rikka0w0: fixed

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ModifiableRegistry.java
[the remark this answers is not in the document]
coderbot16: I'm not sure if modifiable is the best name, because in this case it's referring more to the ability to clear or remove from the registry.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -160,6 +176,23 @@ public void register(V value) {
...
+
+			BiMap<String, StructureFeature<?>> map = StructureFeature.STRUCTURES;
+
+			if (this.oldValue != null && this.oldValue instanceof StructureFeature) {
+				String oldName = ((StructureFeature) oldValue).getName();
+				map.remove(oldName.toLowerCase(Locale.ROOT));
+			}
+
+			map.put(key, structure);
+		}
```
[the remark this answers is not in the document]
rikka0w0: ForgeRegistry.register() is for forge mod only, vanilla and fabric won't use this function. I do agree that we should make our own callbacks.

patchwork-registries/README.md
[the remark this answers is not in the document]
rikka0w0: fixed

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ModifiableRegistry.java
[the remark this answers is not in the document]
rikka0w0: EditableRegistry is better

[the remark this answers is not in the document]
coderbot16: Editable seems like a synonym of modifiable, perhaps ClearableRegistry or RemovableRegistry would work.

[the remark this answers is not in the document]
rikka0w0: forge or vanilla already have ClearableRegistry, so i think we should use RemovableRegistry

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ForgeModRegistryImpl.java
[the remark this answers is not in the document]
rikka0w0: I use mixin to add a method to the Int2ObjectBiMap, because that seems to be the easier. We need to access some private fields, extending the class won't help. I added only 1 method, no fields. BiMap and BitSet are removed.

TheGlitch76:
This is going to be a beast to read through for a final pass but I'll try to get it done within the next day or two.

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/RemovableInt2ObjectBiMap.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.patchworkmc.impl.registries;
+
+public interface RemovableInt2ObjectBiMap<K> {
+	int patchwork_remove(K object);
```
TheGlitch76, line 23: Use a `$` instead of a `_` here: it's the style that's recommended by mixin

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/VanillaRegistry.java
```diff
@@ -0,0 +1,26 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.patchworkmc.impl.registries;
+
+import net.minecraftforge.registries.ForgeRegistry;
+
+public interface VanillaRegistry extends ForgeRegistryProvider {
+	boolean setForgeRegistry(ForgeRegistry forgeRegistry);
```
TheGlitch76, line 25: I don't think this exists in Forge, please use the `patchwork$` prefix.

patchwork-registries/src/main/java/net/patchworkmc/impl/registries/ForgeRegistryProvider.java
```diff
@@ -0,0 +1,26 @@
...
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.patchworkmc.impl.registries;
+
+import net.minecraftforge.registries.ForgeRegistry;
+
+public interface ForgeRegistryProvider {
+	ForgeRegistry getForgeRegistry();
```
TheGlitch76, line 25: If this isn't a forge-added method please use the `patchwork$` prefix

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
```diff
@@ -20,21 +20,35 @@
...
+import com.google.common.collect.Sets;
+import com.google.common.collect.Sets.SetView;
 
 import net.minecraft.util.Identifier;
+import net.minecraft.util.registry.Registry;
 
-import net.patchworkmc.impl.registries.RegistryClassMapping;
+import net.patchworkmc.impl.registries.ForgeRegistryProvider;
 
+//TODO: unimplemented features: saveToDisk, legacyName, sync, dump
```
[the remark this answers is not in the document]
rikka0w0: We don't need saveToDisk feature, it is used in forge's vanilla replacement. Unlike forge, our registry impl is based on vanilla registry, so we don't need to worry about this. It won't corrupt the world.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -204,4 +315,81 @@ public int hashCode() {
...
+	public V remove(Identifier key) {
+		if (!this.isModifiable) {
+			throw new UnsupportedOperationException("Attempted to remove from a non-modifiable Forge Registry");
+		}
+
+		if (this.isLocked()) {
+			throw new IllegalStateException("Attempted to remove from the registry too late.");
+		}
+
+		// If it is modifiable, it must be a forge mod registry, vanilla registries do not support remove().
```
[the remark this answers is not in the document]
rikka0w0: In forge, mod registries support remove (optional), vanilla registries only support insertion and replacement. In our impl, forge mod registries have the same behavior, but vanilla registries only support insertion not replacement. I have a note in the README.md.

patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
```diff
@@ -63,7 +64,10 @@ private static void dispatch(Map<ForgeInitializer, FMLModContainer> mods, Functi
 		for (FMLModContainer container : mods.values()) {
 			ModLoadingContext.get().setActiveContainer(container, new FMLJavaModLoadingContext(container));
 
-			container.getEventBus().post(provider.apply(container));
+			Event event = provider.apply(container);
```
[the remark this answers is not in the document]
rikka0w0: fixed

TheGlitch76:
friendly reminder about a merge conflict, don't worry about bumping module version more.
Will give this a final review later:tm:

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -39,23 +39,98 @@
...
+			this.vanilla = vanilla;
+			((VanillaRegistry) this.vanilla).patchwork$setForgeRegistry(this);
+			this.isVanilla = true;
+
+			// Set the slave map for compatibility
+			this.setSlaveMap(new Identifier("forge", "registry_defaulted_wrapper"), vanilla);
+		}
+
+		// Fabric hooks
+		// TODO: Some vanilla registry types are not patched yet, add this check to avoid crash
```
TheGlitch76, line 110:
```suggestion
		// TODO: Some vanilla registry types are not patched yet, this check is added to avoid a crash
```

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
```diff
@@ -39,23 +39,98 @@
...
-	private Registry<V> vanilla;
-	private Class<V> superType;
-
-	public ForgeRegistry(Identifier name, Registry<V> vanilla, Class<V> superType) {
+	private final Identifier name; // The forge name
+	private final boolean isVanilla;
+	private final Registry<V> vanilla;
+	private final Class<V> superType;
+	private final Map<Identifier, ?> slaves = new HashMap<>();
+	private final CreateCallback<V> create;
```
TheGlitch76, line 60: all these callbacks could use a `Callback` suffix, so `create` -> `createCallback` etc. It makes the code a bit easier to follow

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryBuilder.java
```diff
@@ -0,0 +1,352 @@
...
+			this.add((BakeCallback<T>) inst);
+		}
+
+		if (inst instanceof DummyFactory) {
+			this.set((DummyFactory<T>) inst);
+		}
+
+		if (inst instanceof MissingFactory) {
+			this.set((MissingFactory<T>) inst);
+		}
```
TheGlitch76, line 140:
```suggestion
		if (inst instanceof AddCallback) {
			this.add((AddCallback<T>) inst);
		} else if (inst instanceof ClearCallback) {
			this.add((ClearCallback<T>) inst);
		} else if (inst instanceof CreateCallback) {
			this.add((CreateCallback<T>) inst);
		} else if (inst instanceof ValidateCallback) {
			this.add((ValidateCallback<T>) inst);
		} else if (inst instanceof BakeCallback) {
			this.add((BakeCallback<T>) inst);
		} else if (inst instanceof DummyFactory) {
			this.set((DummyFactory<T>) inst);
		} else if (inst instanceof MissingFactory) {
			this.set((MissingFactory<T>) inst);
		}
```

patchwork-registries/src/main/java/net/minecraftforge/registries/RegistryManager.java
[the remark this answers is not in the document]
TheGlitch76: That includes modded registries?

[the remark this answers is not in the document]
rikka0w0: Yes.
The vanilla registry handles the saveToDisk, and fabric handles sync.
legacyName allows forge mods to get the IForgeRegistry by name, now it seems to be deprecated, the only way for forge mod to obtain a IForgeRegistry instance is to call GameRegistry.findRegistry(<Class<?>).
dump just prints the registry entry and its id.

patchwork-registries/src/main/java/net/minecraftforge/registries/ForgeRegistry.java
rikka0w0: fixed

TheGlitch76:
Please fix the merge conflict

rikka0w0:
Now it should be ready to be merged

PR #82, opened 2020-06-15
Add the ClientTickEvent event

ZNixian:
Implements the ClientTickEvent event.

I've verified the generated bytecode from the mixin, should behave identically to [Forge](https://github.com/PatchworkMC/YarnForge/blob/04d384a/patches/minecraft/net/minecraft/client/MinecraftClient.java.patch#L247).


ZNixian:
This will need squashing when merged due to the formatting fix commit

PR #87, opened 2020-06-16
Add more missing function to ModList, FMLModContainer and ModContainer

rikka0w0:
FMLModContainer and ModContainer:
 [API] Add getMod and acceptEvent methods
 [Internal] setParent and getParent: associated Fabric ModContainer
getter and setters

ModList:
 Add most helper functions present in Forge's ModList, these are
frequenty used by mods

ModLoader(class):
 A class may used by mods to post event on the FML event bus

TODO:
the ModContainer.setMod(Object) is not called anywhere in the code, I don't know how the forge mod main class instance is created. This is the only problem.

### DIFF
--- a/patchwork-fml/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
+++ b/patchwork-fml/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
@@ -27,6 +27,7 @@ import net.minecraftforge.fml.ModContainer;
 
 public class FMLModContainer extends ModContainer {
 	private final IEventBus eventBus;
+	private Object instance;
 
 	public FMLModContainer(String id) {
 		super(id);
@@ -46,5 +47,19 @@ public class FMLModContainer extends ModContainer {
 
 	private void onEventFailed(IEventBus iEventBus, Event event, IEventListener[] listeners, int i, Throwable throwable) {
 		// TODO
+	}
+
+	@Override
+	protected void acceptEvent(final Event e) {
+		this.eventBus.post(e);
+	}
+
+	@Override
+	public Object getMod() {
+		return this.instance;
+	}
+
+	public void setMod(Object instance) {
+		this.instance = instance;
 	}
 }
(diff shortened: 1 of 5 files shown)


patchwork-fml/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
```diff
@@ -47,4 +48,18 @@ public IEventBus getEventBus() {
...
+	@Override
+	protected void acceptEvent(final Event e) {
+		this.eventBus.post(e);
+	}
+
+	@Override
+	public Object getMod() {
+		return this.instance;
+	}
+
```
TheGlitch76, line 61:
```suggestion

	// TODO: This isn't called anywhere yet
```

patchwork-fml/src/main/java/net/minecraftforge/fml/javafmlmod/FMLModContainer.java
```diff
@@ -47,4 +48,18 @@ public IEventBus getEventBus() {
...
 	}
+
+	@Override
+	protected void acceptEvent(final Event e) {
+		this.eventBus.post(e);
+	}
+
+	@Override
+	public Object getMod() {
+		return this.instance;
```
TheGlitch76, line 59: this seems to always be null, it might actually be better to drop them until the setter is actually called

rikka0w0: Should we just comment getMod() and setMod() for now? Impl setMod needs some changes in the patcher as well, the ForgeInitializer.onForgeInitialize should return the instance of the Forge mod class, so that we can record it in its ModContainer.

TheGlitch76: yeah, just comment them out for now and the change that needs Patcher can be a separate PR

rikka0w0: Can we leave getMod() and setMod() here? Just add TODOs to both of them and throw a warning when they are called? Other methods in this pr require getMod(), simply remove getMod() needs more work.

TheGlitch76: Alright. Make sure you open an issue on Patcher about it

Issue #84, opened 2020-06-16
Required features for Curios


ZNixian:
[CurseForge](https://www.curseforge.com/minecraft/mc-mods/curios)

`patchwork-gui`
- [ ] SlotItemHandler
- [ ] ItemStackHandler
- [ ] IItemHandler
- [ ] IContainerFactory
- [ ] IForgeContainerFactory
- [ ] ItemTooltipEvent

`patchwork-events-entity`
- [ ] LivingDropsEvent

`patchwork-events-lifecycle` (probably the best place for this):
- [ ] ServerLifecycleHooks.getCurrentServer

Unknown module:
- [ ] EntityRenderDispatcher.getSkinMap must be added

Bugs:
* `FMLLoadCompleteEvent` must be sent after `MinecraftClient.init` has returned, since Curios uses fields from `MinecraftClient`.

Access widening:
* Screens.Provider and Screens.register must be made accessible

TheGlitch76:
I can handle the Screens AWs and the ContainerFactories. I've got patchwork-containers started locally but it needed to be rewritten since Nuclear was using MethodHandles from pre-AW.

Devan-Kerman:
this is on fabric too :tiny_potato:

TheGlitch76:
> This mod is interesting largely because it's required for Botania

:tiny_potato:

kitlith:
In addition, curios is available on fabric for 1.16, but patchwork is still on 1.14.

There are several mod APIs at this point that I think it'd be worth investigating forge mods interacting with the fabric version of the api, but the ones I've seen so far are not present on 1.14, which makes testing that and developing a framework around that a bit more difficult.

In the meantime, there are always more forge events and stuff to implement.

florensie:
Curios API uses IMC on Forge so using the Fabric counterpart wouldn't work here.

TheGlitch76:
Hypothetically someone would make a wrapper lib that translates the IMC API to the underlying fabric one, but that's out of scope for here.

PR #89, opened 2020-06-17
Add the PlayerXpEvent events (add XP points/level, pickup XP orb)

ZNixian:
Required for Curios, See #84.

I can't remember if it is in Forge, but I've marked `PlayerPickupXpEvent` as deprecated as it's removed in Forge for 1.15. It seems to have been the original event for interacting with XP pickups, until it was superseded by the `PlayerXpEvent` family of events.

When `PlayerXpEvent.XpChange` is cancelled, there is a very small behavioural difference to Forge: `PlayerEntity.addScore` (with zero as an argument) and `PlayerEntity.getNextLevelExperience` are called. The former calls `get` and `set` on a DataTracker, the latter doesn't call any other methods (and while other mods hooking `getNextLevelExperience` is plausible, one would expect those not to have any side-effects).

Since a `@ModifyVariable` mixin is used, it's not (as far as I know) straightforward to also properly cancel this event, though of course it can be done if you think it might make a difference.

### DIFF
--- a/patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/player/PlayerXpEvent.java
+++ b/patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/player/PlayerXpEvent.java
@@ -1,0 +1,90 @@
+/*
+ * Minecraft Forge, Patchwork Project
+ * Copyright (c) 2016-2020, 2019-2020
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.event.entity.player;
+
+import net.minecraft.entity.ExperienceOrbEntity;
+import net.minecraft.entity.player.PlayerEntity;
+
+public abstract class PlayerXpEvent extends PlayerEvent {
+	public PlayerXpEvent(PlayerEntity player) {
+		super(player);
+	}
+
+	// For legacy reasons, the instances of this class should actually be of
+	// type PlayerPickupXpEvent
+	public static class PickupXp extends CancelablePlayerXpEvent {
+		private final ExperienceOrbEntity orb;
+
+		public PickupXp(PlayerEntity player, ExperienceOrbEntity orb) {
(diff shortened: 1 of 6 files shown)


patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/player/PlayerXpEvent.java
```diff
@@ -0,0 +1,90 @@
...
+import net.minecraft.entity.ExperienceOrbEntity;
+import net.minecraft.entity.player.PlayerEntity;
+
+public abstract class PlayerXpEvent extends PlayerEvent {
+	public PlayerXpEvent(PlayerEntity player) {
+		super(player);
+	}
+
+	// For legacy reasons, the instances of this class should actually be of
+	// type PlayerPickupXpEvent
```
TheGlitch76, line 31: This should be a javadoc.

patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinExperienceOrbEntity.java
```diff
@@ -0,0 +1,47 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.entity.ExperienceOrbEntity;
+
+@Mixin(ExperienceOrbEntity.class)
+public class MixinExperienceOrbEntity {
+	// After checking we're on the server and the player is ready to pick up the orb, the first
+	// thing the target method does is set experiencePickUpDelay, hence hook just before that.
+	@Inject(method = "onPlayerCollision", cancellable = true, at = @At(value = "FIELD", opcode = Opcodes.H_PUTFIELD, ordinal = 0,
```
TheGlitch76, line 37: This Handle opcode looks suspicious. Have you tested this mixin and confirmed it works?

TheGlitch76: While you're at it please make sure this launches in a reobf environment (compiled and in MultiMC/vanilla launcher)

ZNixian: I've tested it in a dev environment, I'll test it in MultiMC now.

famous1622: Looking at the patch this seems to be identical.

ZNixian: As in the commit I just pushed? That converted some comments to a JavaDoc comment

ZNixian: Clarified over discord that was referring to the bytecode matching.

ZNixian: Since I didn't set it up properly I thought it wasn't working in reobf, tested it now and it works fine as expected.

PR #90, opened 2020-06-19
Implement FMLServerAboutToStartEvent and hooks related to it.

famous1622:
This PR implements FMLServerAboutToStartEvent, and some other calls that happen in forge at the same time (mostly related to passing the MinecraftServer object around). It also creates a stub version of ServerLifecycleHooks that only implements getCurrentServer, which completes part of #84

### DIFF
--- a/patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
+++ b/patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
@@ -32,7 +32,7 @@ import net.patchworkmc.impl.event.lifecycle.LifecycleEvents;
 @Mixin(MinecraftDedicatedServer.class)
 public class MixinMinecraftDedicatedServer {
 	@Inject(method = "setupServer", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/UserCache;setUseRemote(Z)V", shift = At.Shift.AFTER))
-	private void onServerAboutToStart(CallbackInfoReturnable<Boolean> cir) {
+	private void onServerAboutToStart(CallbackInfo ci) {
 		LifecycleEvents.handleServerAboutToStart((MinecraftServer) (Object) this);
 	}
 }
--- a/patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
+++ b/patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
@@ -1,0 +1,38 @@
+/*
+ * Minecraft Forge, Patchwork Project
+ * Copyright (c) 2016-2020, 2019-2020
+ *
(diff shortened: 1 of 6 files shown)


patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
```diff
@@ -0,0 +1,38 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.server.dedicated.MinecraftDedicatedServer;
+
+import net.patchworkmc.impl.event.lifecycle.LifecycleEvents;
+
+@Mixin(MinecraftDedicatedServer.class)
+public class MixinMinecraftDedicatedServer {
+	@Inject(method = "setupServer", at = @At(value = "INVOKE", target = "Lnet/minecraft/util/UserCache;setUseRemote(Z)V", shift = At.Shift.AFTER))
```
TheGlitch76, line 34: This is misleadingly named but it's the injection point for "after a method call"
```suggestion
	@Inject(method = "setupServer", at = @At(value = "INVOKE_ASSIGN", target = "Lnet/minecraft/util/UserCache;setUseRemote(Z)V"))
```

patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
```diff
@@ -0,0 +1,38 @@
...
+	private void onServerAboutToStart(CallbackInfoReturnable<Boolean> cir) {
```
TheGlitch76, line 35: Use a normal CallbackInfo for this

patchwork-events-lifecycle/src/main/java/net/patchworkmc/mixin/event/lifecycle/MixinMinecraftDedicatedServer.java
TheGlitch76: Actually I'll have to check this because modders say one thing and mixin docs say another

TheGlitch76: How it is now is fine.

patchwork-events-lifecycle/src/main/resources/patchwork-events-lifecycle.mixins.json
```diff
@@ -11,6 +11,9 @@
     "MixinClientWorld",
     "MixinMinecraftClient"
```
ZNixian, line 12: Sorry, missed the boat on this PR. Shouldn't MixinIntegratedServer be included here? I can't see it in any of the Mixin JSON files.

TheGlitch76: Good catch! I will fix right now.

PR #93, opened 2020-06-20
add a missed mixin

TheGlitch76:
thanks to @ZNixian for the catch


famous1622:
Was my bad, just tested and it works fine.

ramidzkh:
A clear example of the law of triviality. Important details get overlooked

Devan-Kerman:
🚲🏠

PR #96, opened 2020-06-21
Implement the LivingDropsEvent event

ZNixian:
Required for Curios, see #84.

This PR is largely comprised of two parts:

Firstly, it implements `Entity.captureDrops` in the `patchwork-extensions` module. This is a collection that can be set on an entity that will redirect all the `ItemEntity`s it would otherwise drop on the ground into said collection.

The second part (in the second commit) adds to `LivingEntity` to use this mechanism to catch all the items the LivingEntity drops when it dies, and pass it through an event to let mods alter the dropped items. Finally it spawns them if the event wasn't cancelled.

Note that the second part creates a dependency from `patchwork-events-entity` to `patchwork-extensions`, however it seemed like the most logical place to put both parts.

The second part does involve some hackery to access the looting level after the local variable has been dropped, passing it through a field (described in a JavaDoc comment). I could just call `EnchantmentHelper.getLooting` again, but we'd have to use this method anyway when we implement `LootingLevelEvent`.

### DIFF
--- a/patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/living/LivingDropsEvent.java
+++ b/patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/living/LivingDropsEvent.java
@@ -1,0 +1,83 @@
+/*
+ * Minecraft Forge, Patchwork Project
+ * Copyright (c) 2016-2020, 2019-2020
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.event.entity.living;
+
+import java.util.Collection;
+
+import net.minecraftforge.common.MinecraftForge;
+
+import net.minecraft.entity.ItemEntity;
+import net.minecraft.entity.LivingEntity;
+import net.minecraft.entity.damage.DamageSource;
+
+/**
+ * LivingDropsEvent is fired when an Entity's death causes dropped items to appear.
+ *
+ * <p>This event is fired whenever an Entity dies and drops items in {@link LivingEntity#onDeath(DamageSource)}.</p>
+ *
+ * <p>
+ * {@link #source} contains the DamageSource that caused the drop to occur.<br>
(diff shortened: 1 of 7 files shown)


patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinLivingEntity.java
```diff
@@ -101,6 +115,65 @@ private void hookHandleFallDamageCancel(float distance, float damageMultiplier,
 		}
 	}
 
+	/**
+	 * Carry over the looting level between the two mixins for the drop method.
```
TheGlitch76, line 119: Thank you for the descriptive javadoc!

patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/living/LivingDropsEvent.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * LivingDropsEvent is fired when an Entity's death causes dropped items to appear.
+ *
+ * <p>This event is fired whenever an Entity dies and drops items in {@link LivingEntity#onDeath(DamageSource)}.</p>
+ *
+ * <p>
+ * {@link #source} contains the DamageSource that caused the drop to occur.<br>
+ * {@link #drops} contains the ArrayList of EntityItems that will be dropped.<br>
+ * {@link #lootingLevel} contains the amount of loot that will be dropped.<br>
+ * {@link #recentlyHit} determines whether the Entity doing the drop has recently been damaged.<br>
```
Nuclearfarts, line 39: Forge javadoc once again proves to be awful. A few minor nitpicks and a missed MCP name here:
```suggestion
 * {@link #source} contains the {@link DamageSource} that caused the drop to occur.<br>
 * {@link #drops} contains the Collection of {@link ItemEntity}s that will be dropped.<br>
 * {@link #lootingLevel} contains the level of Looting used to kill the entity doing the drop.<br>
 * {@link #recentlyHit} determines whether the entity doing the drop has recently been damaged by a player or tamed wolf.<br>
```

patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinLivingEntity.java
```diff
@@ -101,6 +115,65 @@ private void hookHandleFallDamageCancel(float distance, float damageMultiplier,
...
+
+	@Inject(method = "drop", at = @At(value = "FIELD", target = "Lnet/minecraft/entity/LivingEntity;playerHitTimer : I"), locals = LocalCapture.CAPTURE_FAILHARD)
+	private void hookDropForCapturePre(DamageSource src, CallbackInfo info, int lootingLevel) {
+		IForgeEntity forgeEntity = (IForgeEntity) this;
+		forgeEntity.captureDrops(new ArrayList<>());
+
+		// Make sure no other threads are currently in this method first
+		if (!dropLevelGuard.compareAndSet(null, Thread.currentThread())) {
+			throw new ConcurrentModificationException("LivingEntity.drop: Single-threaded local hack used by multiple threads");
+		}
```
Nuclearfarts, line 155: This is an awful hack -- unless you have a good reason to not, you should just use `ThreadLocal`.

PR #97, opened 2020-06-21
Make stack accessible for mods

ramidzkh:
BoP uses it and so does @valoeghese's LGT test mod


famous1622:
It matches with the bot, so LGTM

valoeghese:
Why is this not yet merged
it's been tested and proved well
and thus, as I have urged
you merge this, or the bell shall knell
And thus sing the people:

"The API functions not,
Yet it is only one line off
Seems the triage team cannot
do their job - what a turnoff"
And thus sing the people.

Nuclearfarts:
My only concern here is that we should probably do it with a mixin, but given that mods use it it's probably fine.

PR #101, opened 2020-06-22
EntityItemPickupEvent + PlayerEvent#ItemPickupEvent

Rongmario:
Used `@Overwrite` in latest commit as injections and redirects got crazy.

Roughly checked, and everything should be working.

These events are so similarly named, don't blame me for the EntityEvents methods, if you have any good ones pls say.


TheGlitch76:
:concern:

patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
```diff
@@ -0,0 +1,95 @@
...
+	 *
+	 * <p>Small changes to Forge's implementation:
+	 * 	  1. pickupDelay is only checked once.
+	 * 	  2. Variables are assigned after the the first return.</p>
+	 */
+	@Overwrite
+	public void onPlayerCollision(PlayerEntity player) {
+		if (!this.world.isClient && this.pickupDelay <= 0) {
+			final ItemEntity entity = (ItemEntity) (Object) this;
+			int result = EntityEvents.onItemPickup(entity, player);
```
ZNixian, line 70: Would it be easier just to move the event stuff in here? Without checking two different files it's not obvious what the result represents - `event.getResult() == Result.ALLOW` is much clearer than `result == 1`?

patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
```diff
@@ -0,0 +1,95 @@
...
+	 * 3 injections and 2 redirects is needed. Which at that point, Overwrite might be a better option.</p>
+	 *
+	 * <p>This also eliminates the need of holding 2 references (that might have used ThreadLocal)
+	 * for {@link EntityItemPickupEvent}'s result and a copied ItemStack.</p>
+	 *
+	 * <p>Small changes to Forge's implementation:
+	 * 	  1. pickupDelay is only checked once.
+	 * 	  2. Variables are assigned after the the first return.</p>
+	 */
+	@Overwrite
```
TheGlitch76, line 66: I'm split on using an Overwrite here.
On one hand, overwrite bad and it could cause some problems with other mods.
On the other hand, this is potentially hot code and the mixins were a bit nasty (but still tame).
I'm going to sleep on this tonight and think about if this is the right solution.

patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
Rongmario: It will be easier to move from EntityEvents back to its respective Hook class to prevent breaking compatibility with mods that use the provided hooks to fire the event.

TheGlitch76:
On further review I'd like the threadlocal version to be used until we have a better alternative.

Rongmario:
Sure I'll revert to the ThreadLocal commit.

Rongmario:
Just updated, reverted Overwrite, and implemented rest of the entity.item events. More info see latest commit msg.

patchwork-events-entity/src/main/java/net/minecraftforge/event/entity/item/ItemTossEvent.java
```diff
@@ -0,0 +1,52 @@
...
+import net.minecraft.entity.player.PlayerEntity;
+
+/**
+ * Event that is fired whenever a player tosses (Q) an item or drag-n-drops a
+ * stack of items outside the inventory GUI screens. Canceling the event will
+ * stop the items from entering the world, but will not prevent them being
+ * removed from the inventory - and thus removed from the system.
+ */
+// TODO: Call Location - ForgeHook#onPlayerTossEvent -> PlayerEntity#dropSelectedItem + PlayerEntity#dropItem
+public class ItemTossEvent extends ItemEvent {
```
TheGlitch76, line 32: May as well leave this out if you're not going to implement the hooks for it

Rongmario: I can do it in the same commit if its cool :^)

TheGlitch76: Go ahead

patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	@ModifyConstant(method = "onPlayerCollision", constant = @Constant(intValue = 6000, ordinal = 0))
+	private int onPlayerCollideCheckLifespan(int original) {
+		return lifespan;
+	}
+
+	@Redirect(method = "onPlayerCollision", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/player/PlayerEntity;sendPickup(Lnet/minecraft/entity/Entity;I)V"))
+	private void onPlayerItemPickup(PlayerEntity entity, Entity item, int count) {
+		EntityEvents.firePlayerItemPickupEvent(entity, (ItemEntity) item, getStack().copy());
```
rikka0w0, line 118: I also implemented this event in #118.
`patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java`

I created `PlayerEvents` to fire all types of PlayerEvent, so that `EntityEvents` can be less messy. I would suggest one of us should drop his `PlayerEvent.ItemPickupEvent`  implementation. Sorry for the inconvenience, I didn't know this PR before I started #118.

Something I would like to discuss is that, in 1.14.4 (The yarnforge snapshot), Forge removed the `entityIn.sendPickup(this, i);` before `if (itemstack.isEmpty()) {` but in the later version they didnt. However, I don't think firing the PlayerEvent.ItemPickupEvent before or after `entityIn.sendPickup(this, i);` make any noticable difference.

TheGlitch76:
Merge conflicts

Rongmario:
Deprecating for: https://github.com/PatchworkMC/patchwork-api/pull/118. I'll draft a tiny PR for EntityItemPickupEvent. And ItemTossEvent will have to wait until https://github.com/PatchworkMC/patchwork-api/pull/118 is PR'd.

PR #109, opened 2020-07-05
Add ExtensionPoint to prevent one-sided mods from crashing

rikka0w0:
In 1.14, Forge removed “sidedness” attribute from the `mods.toml` and now one-side mods have to call this in their mod constructor to skip side checking:
```java
//Make sure the mod being absent on the other network side does not cause the client to display the server as incompatible
ModLoadingContext.get().registerExtensionPoint(ExtensionPoint.DISPLAYTEST, () -> Pair.of(() -> FMLNetworkConstants.IGNORESERVERONLY, (a, b) -> true));
```

This pr adds `ExtensionPoint.DISPLAYTEST` and relevent ExtensionPoint functions. However, due to a limitation of the Fabric Loader, Patchwork will not check for the presence of the Forge mod on the other side.

### DIFF
--- a/patchwork-fml/src/main/java/net/minecraftforge/fml/ExtensionPoint.java
+++ b/patchwork-fml/src/main/java/net/minecraftforge/fml/ExtensionPoint.java
@@ -1,0 +1,55 @@
+/*
+ * Minecraft Forge, Patchwork Project
+ * Copyright (c) 2016-2020, 2019-2020
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation version 2.1
+ * of the License.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.minecraftforge.fml;
+
+import java.util.function.BiFunction;
+import java.util.function.BiPredicate;
+import java.util.function.Supplier;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.gui.screen.Screen;
+
+public class ExtensionPoint<T> {
+	public static final ExtensionPoint<BiFunction<MinecraftClient, Screen, Screen>> CONFIGGUIFACTORY = new ExtensionPoint<>();
+	// TODO: Not used by any Forge code, ModFileResourcePack is not implemented in Patchwork API
+	// public static final ExtensionPoint<BiFunction<MinecraftClient, ModFileResourcePack, ResourcePack>> RESOURCEPACK = new ExtensionPoint<>();
+	/**
+	 * Compatibility display test for the mod. Used for displaying compatibility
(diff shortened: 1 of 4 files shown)


patchwork-fml/src/main/java/net/minecraftforge/fml/ModContainer.java
```diff
@@ -46,6 +54,21 @@ public final String getNamespace() {
 		return namespace;
 	}
 
+	@SuppressWarnings("unchecked")
+	public <T> Optional<T> getCustomExtension(ExtensionPoint<T> point) {
+		return Optional.ofNullable((T) extensionPoints.getOrDefault(point, () -> null).get());
```
TheGlitch76, line 59: > getOrDefault(x () -> null)

Isn't that the same thing is `get()`? Maybe this is some quirk of IdentityHashMap but I don't think so.

patchwork-fml/src/main/java/net/minecraftforge/fml/ExtensionPoint.java
```diff
@@ -0,0 +1,55 @@
...
+	/**
+	 * Compatibility display test for the mod. Used for displaying compatibility
+	 * with remote servers with the same mod, and on disk saves.
+	 *
+	 * <p>The supplier provides my "local" version for sending across the network or
+	 * writing to disk The predicate tests the version from a remote instance or
+	 * save for acceptability (Boolean is true for network, false for local save)
+	 *
+	 * <p>TODO: Fabric servers do not check for client's mod list,
+	 * there is no way to implement the DISPLAYTEST function in Patchwork.
```
TheGlitch76, line 44: There's no way yet--we will eventually have to layer our own networking ontop of what Fabric has, but that's waiting on some work to be done by liach and co.

patchwork-fml/src/main/java/net/minecraftforge/fml/ExtensionPoint.java
```diff
@@ -0,0 +1,55 @@
...
+public class ExtensionPoint<T> {
+	public static final ExtensionPoint<BiFunction<MinecraftClient, Screen, Screen>> CONFIGGUIFACTORY = new ExtensionPoint<>();
+	// TODO: Not used by any Forge code, ModFileResourcePack is not implemented in Patchwork API
+	// public static final ExtensionPoint<BiFunction<MinecraftClient, ModFileResourcePack, ResourcePack>> RESOURCEPACK = new ExtensionPoint<>();
+	/**
+	 * Compatibility display test for the mod. Used for displaying compatibility
+	 * with remote servers with the same mod, and on disk saves.
+	 *
+	 * <p>The supplier provides my "local" version for sending across the network or
+	 * writing to disk The predicate tests the version from a remote instance or
```
TheGlitch76, line 40:
```suggestion
	 * <p>The supplier provides the "local" version for sending across the network or
	 * writing to disk. The predicate tests the version from a remote instance or
```

patchwork-fml/src/main/java/net/minecraftforge/fml/ModContainer.java
rikka0w0: Maybe Forge was trying to provide some fallback so they leave `() -> null` here.

Actually I'm not too sure either, perhaps we can leave it there to keep the behavior consistent with Forge.

patchwork-fml/src/main/java/net/minecraftforge/fml/ExtensionPoint.java
rikka0w0: I agree, so for now we just keep these dummy extension points to avoid `ClassNotFoundException`.

patchwork-fml/src/main/java/net/minecraftforge/fml/ModContainer.java
TheGlitch76: Hang on, i see what's going on. It's returning some supplier, and this is more concise because it avoids having to check before unboxing.

PR #112, opened 2020-07-08
Make Biomes O' Plenty work

### LINKED ISSUE #98: Required Features for Biomes O Plenty
Please let me know any others!
What I know so far is that we need this

- [x] IPlantable
image
- [x] Tool Types (#73 does this stage)
image
- [ ] Block.Settings harvest level, which is what bop uses tool types for
image

leo60228:
Closes #98 
Requires PatchworkMC/patchwork-patcher#60 (merged)
Partially blocked on PatchworkMC/patchwork-patcher#57

### DIFF
--- a/patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
+++ b/patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
@@ -19,6 +19,7 @@
 
 package net.minecraftforge.common;
 
+import java.util.Collection;
 import java.util.Objects;
 import java.util.ArrayList;
 import java.util.List;
@@ -84,11 +85,11 @@ public class BiomeManager {
 	}
 
 	// Biomes O' Plenty pokes Forge internals. Fun
-	private static ArrayList<BiomeEntry>[] biomes = setupBiomes();
+	private static TrackedList<BiomeEntry>[] biomes = setupBiomes();
 
-	private static ArrayList<BiomeEntry>[] setupBiomes() {
+	private static TrackedList<BiomeEntry>[] setupBiomes() {
 		@SuppressWarnings("unchecked")
-		ArrayList<BiomeEntry>[] currentBiomes = new ArrayList[BiomeType.values().length];
+		TrackedList<BiomeEntry>[] currentBiomes = new TrackedList[BiomeType.values().length];
 		List<BiomeEntry> list = new ArrayList<BiomeEntry>();
 
 		list.add(new BiomeEntry(Biomes.FOREST, 10));
@@ -98,7 +99,7 @@ public class BiomeManager {
 		list.add(new BiomeEntry(Biomes.BIRCH_FOREST, 10));
 		list.add(new BiomeEntry(Biomes.SWAMP, 10));
 
-		currentBiomes[BiomeType.WARM.ordinal()] = new ArrayList<BiomeEntry>(list);
+		currentBiomes[BiomeType.WARM.ordinal()] = new TrackedList<BiomeEntry>(list);
 		list.clear();
 
 		list.add(new BiomeEntry(Biomes.FOREST, 10));
@@ -106,16 +107,16 @@ public class BiomeManager {
 		list.add(new BiomeEntry(Biomes.TAIGA, 10));
 		list.add(new BiomeEntry(Biomes.PLAINS, 10));
 
-		currentBiomes[BiomeType.COOL.ordinal()] = new ArrayList<BiomeEntry>(list);
+		currentBiomes[BiomeType.COOL.ordinal()] = new TrackedList<BiomeEntry>(list);
 		list.clear();
 
 		list.add(new BiomeEntry(Biomes.SNOWY_TUNDRA, 30));
 		list.add(new BiomeEntry(Biomes.SNOWY_TAIGA, 10));
 
-		currentBiomes[BiomeType.ICY.ordinal()] = new ArrayList<BiomeEntry>(list);
+		currentBiomes[BiomeType.ICY.ordinal()] = new TrackedList<BiomeEntry>(list);
 		list.clear();
 
-		currentBiomes[BiomeType.DESERT.ordinal()] = new ArrayList<BiomeEntry>(list);
+		currentBiomes[BiomeType.DESERT.ordinal()] = new TrackedList<BiomeEntry>(list);
 
 		return currentBiomes;
 	}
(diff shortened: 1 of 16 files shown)


patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -64,14 +69,61 @@ private OverworldClimate getClimate() {
...
 			this.biome = biome;
 		}
 
 		private int getWeight() {
 			return this.weight;
 		}
 	}
+
+	// Biomes O' Plenty pokes Forge internals. Fun
+	private static ArrayList<BiomeEntry>[] biomes = setupBiomes();
```
famous1622, line 87: Best practice would be to type this as List, rather than ArrayList, we shouldn't care about the class of the object behind it.

patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -64,14 +69,61 @@ private OverworldClimate getClimate() {
...
+
+	private static ArrayList<BiomeEntry>[] setupBiomes() {
```
famous1622, line 89: Same here.

patchwork-extensions/src/main/java/net/patchworkmc/mixin/extension/MixinMinecraftClient.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.color.item.ItemColors;
+
+@Mixin(MinecraftClient.class)
+public class MixinMinecraftClient {
+	@Shadow
+	private ItemColors itemColorMap;
+
+	public ItemColors getItemColors() {
+		return this.itemColorMap;
```
famous1622, line 34: Could this just be implemented as an Accessor?

patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
famous1622: Wait does Forge type this as ArrayList? If so :concern:

Rongmario: IIRC it was a custom implementation of ArrayList to track if the ArrayList had been modified

leo60228: BOP calls setupBiomes directly through reflection. It casts the result to a List, but since the actual Forge type extends ArrayList, I thought this would be better.

famous1622: Why not go all the way and implement the class from forge in case someone casts to it?

leo60228: True, that just didn't seem necessary. I'll do that later.

patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -64,14 +70,134 @@ private OverworldClimate getClimate() {
...
+		list.clear();
+
+		currentBiomes[BiomeType.DESERT.ordinal()] = new TrackedList<BiomeEntry>(list);
+
+		return currentBiomes;
+	}
+
+	public static ImmutableList<BiomeEntry> getBiomes(BiomeType type) {
+		int idx = type.ordinal();
+		List<BiomeEntry> list = idx >= biomes.length ? null : biomes[idx];
```
TheGlitch76, line 126: this line makes sense but it's difficult to tell what it's doing at first glance
```suggestion
		// We want to return null instead of throw an exception if the index is out of bounds.
		List<BiomeEntry> list = idx >= biomes.length ? null : biomes[idx];
```

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/MixinBlockSettings.java
```diff
@@ -0,0 +1,80 @@
...
+
+import net.minecraft.block.Block;
+import net.minecraft.item.Item;
+import net.minecraft.tag.Tag;
+
+import net.fabricmc.fabric.api.block.FabricBlockSettings;
+import net.fabricmc.fabric.api.tools.FabricToolTags;
+
+@Mixin(Block.Settings.class)
+public class MixinBlockSettings {
```
TheGlitch76, line 34: This needs an interface so that the added methods can be accessed by fabric mods.

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/MixinBlockSettings.java
```diff
@@ -0,0 +1,80 @@
...
+			break;
+		case "shovel":
+			this.miningTool = FabricToolTags.SHOVELS;
+			break;
+		case "sword":
+			this.miningTool = FabricToolTags.SWORDS;
+			break;
+		}
+
+		FabricBlockSettings fabric = FabricBlockSettings.copyOf((Block.Settings) (Object) this);
```
TheGlitch76, line 72: Thanks for using the fabric version here!

patchwork-extensions-block/src/main/resources/patchwork-extensions-block.accesswidener
```diff
@@ -0,0 +1,9 @@
+accessWidener v1 named
+
+accessible method net/minecraft/block/StairsBlock <init> (Lnet/minecraft/block/BlockState;Lnet/minecraft/block/Block$Settings;)V
+accessible field net/minecraft/block/Block materialColor Lnet/minecraft/block/MaterialColor;
+accessible method net/minecraft/block/DoorBlock <init> (Lnet/minecraft/block/Block$Settings;)V
+accessible method net/minecraft/block/TrapdoorBlock <init> (Lnet/minecraft/block/Block$Settings;)V
+accessible method net/minecraft/class_2440 <init> (Lnet/minecraft/class_2440$class_2441;Lnet/minecraft/class_2248$class_2251;)V
```
TheGlitch76, line 7: Mind adding a comment about this class's 1.15/1.16 name if it still exists?

patchwork-extensions/src/main/java/net/patchworkmc/mixin/extension/MixinAbstractTreeFeature.java
```diff
@@ -0,0 +1,44 @@
...
+import net.minecraft.world.IWorld;
+import net.minecraft.world.ModifiableTestableWorld;
+import net.minecraft.util.math.BlockPos;
+
+@Mixin(AbstractTreeFeature.class)
+public abstract class MixinAbstractTreeFeature {
+	@Shadow
+	protected abstract void setToDirt(ModifiableTestableWorld world, BlockPos pos);
+
+	protected void setDirtAt(ModifiableTestableWorld reader, BlockPos pos, BlockPos origin) {
```
TheGlitch76, line 36: Not going to worry about it for now, just want a comment on it.
```suggestion
	// TODO: How can we make this accessible in fabric mods?
	protected void setDirtAt(ModifiableTestableWorld reader, BlockPos pos, BlockPos origin) {
```

patchwork-biomes/src/main/java/net/minecraftforge/common/BiomeManager.java
```diff
@@ -64,14 +70,134 @@ private OverworldClimate getClimate() {
 	public static class BiomeEntry extends WeightedPicker.Entry {
 		public final Biome biome;
 
+		public int field_76292_a; // evil hack
```
TheGlitch76, line 73: Can you add a comment on why this is needed?

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/MixinBlockSettings.java
leo60228: This is just a wrapper around Fabric, is there any reason a mod wouldn't just use that directly?

leo60228:
Just rebased.

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/MixinBlockSettings.java
TheGlitch76: source-level patchwork would need this, that kind of transformation would need to be done manually

Issue #21, opened 2020-01-02
Support mods that use ForgeHooks, ForgeEventFactory, etc directly


coderbot16:
Some mods call these functions directly. Patchwork splits them up per module, so we'll need to have a separate module that implements the original Forge versions. Perhaps the module could be `patchwork-event-dispatcher`?

* [ ] [WaMod - MoreBlockVariants](https://www.curseforge.com/minecraft/mc-mods/wamod-more-block-variants)
  * [ ] `ForgeEventFactory#saplingGrowTree`

kitlith:
Outdated issue, this has been implemented (and is even being revisited, heh.)

PR #115, opened 2020-07-09
Implement ForgeHooks, ForgeEventFactory, etc.

kitlith:
Doesn't immediately fix the mod listed in #21 as that specific event either doesn't have a method in EntityEvents or isn't implemented yet. Lays some foundataion I guess.

The PR isn't done yet (gotta check the rest of the events modules at least), but I figured I'd get the bikeshed party started early since I've reached a place where it makes sense to pause.


kitlith:
Alright, I think that covers the trivial redirects. The question now is, how do we want to handle e.g. events that are already implemented but are dispatched directly from mixins instead of through a class? We could either dispatch the event directly from the hook, or we could reorganize those mixins to call into a class that ForgeHooks also delegates to.

TheGlitch76:
Reorg the mixins to call from a PatchworkXEvents, and then call from there.
That should be a separate PR, though.

kitlith:
I'm splitting this into seperate PRs:
 - [x] LifecycleEvents (#128)
 - [ ] EntityEvents
 - [ ] RenderEvents
 - [ ] WorldEvents

PR #116, opened 2020-07-09
Add DrawBlockHighlightEvent(DrawHighlightEvent)

rikka0w0:


patchwork-events-rendering/src/main/java/net/patchworkmc/impl/event/render/RenderEvents.java
```diff
@@ -53,4 +60,23 @@ public static void onTextureStitchPre(SpriteAtlasTexture spriteAtlasTexture, Set
...
+			if (!(target instanceof BlockHitResult)) return false;
+			return MinecraftForge.EVENT_BUS.post(new DrawBlockHighlightEvent.HighlightBlock(context, info, target, subID, partialTicks));
+		case ENTITY:
+			if (!(target instanceof EntityHitResult)) return false;
+			return MinecraftForge.EVENT_BUS.post(new DrawBlockHighlightEvent.HighlightEntity(context, info, target, subID, partialTicks));
+		default:
+			break;
+		}
+
+		return MinecraftForge.EVENT_BUS.post(new DrawBlockHighlightEvent(context, info, target, subID, partialTicks));
```
kitlith, line 80: This return could be moved into the default case instead of being outside the switch block.

kitlith:
Other than nitpicking at forge style, LGTM.

PR #117, opened 2020-07-09
PersistentState(WorldSaveData) implements INBTSerializable

rikka0w0:
Methods in INBTSerializable may be called by Forge mods.


famous1622:
This passes checkstyle, actions just crashed for some reason.

PR #121, opened 2020-07-11
Prepare modules for implementation of ForgeHooks and co.

kitlith:
Required for #115. No, this isn't finished, but I figure getting a PR open before I go sleep makes it easier for other people to comment/review what I'm doing.

a note: Glitch said these classes should be named Patchwork<blah>Events, but I went with <blah>Events as that was the pre-existing style. I can do a commit to refactor these cases into Patchwork<blah>Events though, so lmk if you'd like me to do that.

Each commit should handle a module, and should be able to be looked at individually.

### DIFF
--- a/patchwork-capabilities/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
+++ b/patchwork-capabilities/src/main/java/net/minecraftforge/common/capabilities/CapabilityProvider.java
@@ -23,12 +23,12 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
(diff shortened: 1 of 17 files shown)


kitlith:
okay, that's all the events I found that are called in ForgeHooks, and similar at least. not all of ForgeHooks is sending events though. (i.e. ForgeHooks.canHarvestBlock)

I'm also somewhat dubious about that last commit with patchwork-gui, since those event aren't fired yet, but I decided to go for it since the classes were already present (unlike other methods that I avoided adding because the class was commented out)

kitlith:
I'm splitting this into seperate PRs.

 - [ ] patchwork-
 - [ ] patchwork-
 - [ ] patchwork-
 - [ ] patchwork-

PR #126, opened 2020-07-12
Impl removedByPlayer and canHarvestBlock in IForgeBlock

rikka0w0:
getExpDrops will be always called for all blocks, the exp orb drop behavior follows this rule:
```java
		//			isVanillaBlock		exp		Action
		//	Vanilla		true			xxx		Let the vanilla method do the orb drop
		//	Fabric mod	false			== 0		Let the vanilla method do the orb drop
		//	Forge mod	false			>0		Do orb drop here
		//	Forge mod	false			<=0		Do nothing
```
Fabric mods does not have a method to get the amount of experience dropped from another block. If a Forge mod calls getExpDrop on a Fabric mod block, it will always get 0. This will be changed once Fabric has something equivalent.

__getHarvestTool and getHarvestLevel are stubs__, I will impl more tooltype stuff in another pr.


patchwork-events-world/src/main/java/net/patchworkmc/mixin/event/world/MixinServerPlayerInteractionManager.java
```diff
@@ -50,52 +45,18 @@
...
+		int exp = WorldEvents.onBlockBreakEvent(world, gameMode, player, pos);
 
+		if (exp < 0) {
 			callback.setReturnValue(false);
-		} else if (event.getExpToDrop() != 0) {
-			// TODO: Drop experience
-			throw new UnsupportedOperationException("Cannot drop exp from a BreakEvent yet");
+			callback.cancel();
+		} else {
+			BlockHarvestManager.pushExpDropStack(exp);
```
kitlith, line 54: doing exp stuff in this mixin is a bit confusing when the rest of it is in another mixin in a seperate module.

I understand *why* it's like this, so you don't necessarily need to change it, but I will still say it's a bit confusing.

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/harvest/MixinServerPlayerInteractionManager.java
```diff
@@ -0,0 +1,118 @@
...
+
+		return removed;
+	}
+
+	@Redirect(method = "tryBreakBlock", at = @At(value = "INVOKE", target = Signatures.Block_onBreak, ordinal = 0))
+	private void patchwork$tryBreakBlock_onBreak(Block block, World world, BlockPos pos, BlockState state, PlayerEntity player) {
+		// Suppress this call
+	}
+
+	@Redirect(method = "tryBreakBlock", at = @At(value = "INVOKE", target = Signatures.ServerWorld_removeBlock, ordinal = 0))
```
kitlith, line 68: This is a bit weird, because I'm having trouble finding a "removeBlock" on ServerWorld, let along pointing out where it's called in the yarnforge patch. My best guess is what's called clearBlockState in yarnforge, but I can't find that on ServerWorld either :/

Don't worry about it too much, it's probably something weird in my setup.

rikka0w0: removeBlock is implemented in World not ServerWorld, I believe

rikka0w0:
> The logic here is somewhat tricky, but looks sound.
> 
> If #127 ends up merged before this, can you include the implementation of ForgeHooks.canHarvestBlock?

Yes of course, and also there are other things that I'm going to add to ForgeHooks and ForgeHooksClient. In all of my PRs, the event dispatch is handled by individual classes, and ForgeHooks should only be used for accepting Forge mod calls.

patchwork-events-world/src/main/java/net/patchworkmc/mixin/event/world/MixinServerPlayerInteractionManager.java
rikka0w0: Exp stuff needs the world, block extension and entity package. I cannot think of a better way of doing this.
-1b0d6ec30582ef91

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/harvest/MixinServerPlayerInteractionManager.java
kitlith: Yep, you're right.

patchwork-events-world/src/main/java/net/patchworkmc/impl/event/world/WorldEvents.java
```diff
@@ -57,4 +67,59 @@ public static void onWorldUnload(IWorld world) {
...
+
+			if (entity != null) {
+				BlockEntityUpdateS2CPacket packet = entity.toUpdatePacket();
+
+				if (packet != null) {
+					player.networkHandler.sendPacket(packet);
+				}
+			}
+
+			System.out.println("onBlockBreakEvent cancelled");
```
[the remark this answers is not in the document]
rikka0w0: removed

patchwork-events-world/build.gradle
```diff
@@ -3,5 +3,6 @@ version = getSubprojectVersion(project, "0.3.0")
 
 dependencies {
 	implementation project(path: ':patchwork-api-base', configuration: 'dev')
+	implementation project(path: ':patchwork-extensions-block', configuration: 'dev')
```
TheGlitch76, line 6: Not declared in the fabric.mod.json?

patchwork-extensions-block/build.gradle
```diff
@@ -4,6 +4,7 @@ version = getSubprojectVersion(project, "0.3.0")
 dependencies {
 	implementation project(path: ':patchwork-api-base', configuration: 'dev')
 	implementation project(path: ':patchwork-enum-hacks', configuration: 'dev')
+	implementation project(path: ':patchwork-extensions-item', configuration: 'dev')
```
TheGlitch76, line 7: Not declared in the fabric.mod.json

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/harvest/MixinServerPlayerInteractionManager.java
```diff
@@ -0,0 +1,118 @@
...
+		boolean isCreative = me.isCreative();
+
+		if (isCreative) {
+			patchwork$removeBlock(pos, false);
+		}
+
+		return isCreative;
+	}
+
+	private static final ThreadLocal<Object> tryBreakBlock_canHarvest = BlockContext.createContext(); // flag1
```
TheGlitch76, line 89: @Unique?

patchwork-extensions-block/src/main/resources/patchwork-extensions-block-flammable.mixins.json
```diff
@@ -0,0 +1,11 @@
+{
+  "required": true,
+  "package": "net.patchworkmc.mixin.extensions.block.flammable",
+  "compatibilityLevel": "JAVA_8",
+  "mixins": [
```
TheGlitch76, line 5: This should go in the main `mixins.json` as
`flammable.MixinWorld`

patchwork-extensions-block/src/main/resources/patchwork-extensions-block-harvest.mixins.json
```diff
@@ -0,0 +1,17 @@
+{
+  "required": true,
+  "package": "net.patchworkmc.mixin.extensions.block.harvest",
+  "compatibilityLevel": "JAVA_8",
+  "mixins": [
+    "MixinOreBlock",
```
TheGlitch76, line 6: Up to you on if you want to remove this extra json file too.

rikka0w0: I think we should make separate mixins, otherwise some mixins will get really long. I would like to have separate mixin.json, it is more modular.

TheGlitch76:
Merge conflicts!

PR #127, opened 2020-07-12
Add the patchwork-god-classes module

kitlith:
splitting this out from #115 so that future PRs can implement methods on these classes directly, instead of #115 and/or #121 requiring a rebase + a commit to add new events/methods.

Also means we get to bikeshed the module, documentation, and what classes it's going to contain on its own, rather than bundling that with a bunch of implementation.


kitlith:
A note. patchwork-events-lifecycle has ServerLifecycleHooks, which is in a vein similar to these classes, but I think it can be handled entirely within that module so I don't think it needs to move here.

Similarly, patchwork-networking has NetworkHooks, which I also don't think needs to move here.

kitlith:
renamed to patchwork-god-classes as per discussion in discord, though I'm still open to other names.

PR #131, opened 2020-07-16
Change IForgeRegistryEntry.setRegistryName to return V, not IForgeRegistryEntry

famous1622:
This makes the source and the bytecode return value match with Forge, and is required for CoFHCore to load correctly.


kitlith:
LGTM.

Would it be possible/sane to implement a version that returns V on top of a version that returns IForgeRegistryEntry?

TheGlitch76:
in theory there could be a `patchwork$setRegistryName` that returns `IForgeRegistryEntry` but that doesn't mean we *should*

PR #118, opened 2020-07-10
Impl PlayerChangedDimensionEvent, PlayerRespawnEvent and PlayerLoggedOutEvent

rikka0w0:
Implemented PlayerEvents:
1. ItemPickupEvent
1. ItemCraftedEvent
1. ItemSmeltedEvent
1. PlayerLoggedOutEvent
1. PlayerRespawnEvent
1. PlayerChangedDimensionEvent

The rest of PlayerEvents will be implemented in a separate pr.

PR #134, opened 2020-07-17
Dispatch to EntityEvents from ForgeHooks, etc.

kitlith:
The last of the PRs split off from #115 and #121. This PR may interact with #118 (and others?), as that adds more entity events.


kitlith:
rebased

PR #138, opened 2020-07-21
Clean up stacktrace for unimplemented class / method / field errors.

famous1622:
This keeps stack traces resulting from unimplemented classes and members from being full of confusing Mixin and class loading errors.

### DIFF
--- a/patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
+++ b/patchwork-dispatcher/src/main/java/net/patchworkmc/impl/Patchwork.java
@@ -107,8 +107,31 @@ public class Patchwork implements ModInitializer {
 					error = new PatchworkInitializationException("Failed to construct Patchwork mods");
 				}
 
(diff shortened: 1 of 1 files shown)


rikka0w0:
I would like to suggest that we should include a link for the users to submit their issue, for example, the github issue page.

florensie:
Maybe check if the class/method/field is actually in the forge package?

kitlith:
Forge adds methods and fields to vanilla classes, too, @florensie

florensie:
@kitlith Fair but what if it's something else like a class missing in a dependency mod. I guess it might be too much effort to check every package forge touches.

kitlith:
@florensie we could limit to net.minecraftforge and net.minecraft then? That's still not perfect if a mod adds a class under net.minecraft to access package private things, but it's something.

TheGlitch76:
We don't let mods add things under the net.minecraft package

PR #139, opened 2020-07-22
Better loottable hooks

kitlith:
Opening this as a draft pull request so we can discuss.

I believe there's still an error present, but it's just a matter of how we handle an exception.

The whole idea is to cut out the static `ThreadLocal<Deque<...>>` in LootHooks in favor of messing with json. However, this PR also includes a technique we could use to avoid overwriting the lambda that could be pulled in on its own. (see second commit)


TheGlitch76:
Kit's taking a break from modding, if someone wants to take up this PR they can make PRs to this branch.

kitlith:
I'm still around to answer questions and stuff though ^-^

PR #149, opened 2020-07-24
Use Fabric's world load event.

### LINKED ISSUE #148: Patchwork API incompatible with Fabric API versions ≥0.15.0
As of posting, this incompatibility refers to Fabric API v0.15.0 and v0.15.1 for 1.14.4. Tested against both the release v0.7.0 and commit 494fd271bd27556595545e4f34ecb013127701d3.

Crash log: https://pastebin.com/GxWdzDB1

The problem appears to be a redirect conflict between the new Fabric Lifecycle Events V1 and Patchwork API. 

For reference, here are the conflicting redirects (I believe):

https://github.com/FabricMC/fabric/blob/3673fdf545ad946a8a1cba5e6c5fcdfa88d50900/fabric-lifecycle-events-v1/src/main/java/net/fabricmc/fabric/mixin/event/lifecycle/MinecraftServerMixin.java#L94

https://github.com/PatchworkMC/patchwork-api/blob/494fd271bd27556595545e4f34ecb013127701d3/patchwork-events-world/src/main/java/net/patchworkmc/mixin/event/world/MixinMinecraftServer.java#L80

kitlith:
Fixes #148.

Minor point of discussion: how much do we care about the overworld event being fired earlier than Forge does it? I've separated out the workaround for that into the second commit, so that can easily be removed if we want.

### DIFF
--- a/patchwork-events-world/src/main/java/net/patchworkmc/mixin/event/world/MixinMinecraftServer.java
+++ b/patchwork-events-world/src/main/java/net/patchworkmc/mixin/event/world/MixinMinecraftServer.java
@@ -20,12 +20,8 @@
 package net.patchworkmc.mixin.event.world;
 
 import java.io.IOException;
-import java.util.Iterator;
-import java.util.Map;
 
-import org.spongepowered.asm.mixin.Final;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Redirect;
 
@@ -33,7 +29,6 @@ import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.ServerTask;
 import net.minecraft.server.world.ServerWorld;
 import net.minecraft.util.thread.ReentrantThreadExecutor;
-import net.minecraft.world.dimension.DimensionType;
 
 import net.patchworkmc.impl.event.world.WorldEvents;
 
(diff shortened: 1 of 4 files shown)


patchwork-events-world/src/main/java/net/patchworkmc/mixin/event/world/MixinMinecraftServer.java
```diff
@@ -47,42 +48,10 @@ public MixinMinecraftServer(String name) {
...
-
-	// This injection handles every other dimension type.
-	@Redirect(method = "createWorlds", at = @At(value = "INVOKE", target = "java/util/Map.put (Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;", ordinal = 1))
-	private Object proxyPutWorld(Map<Object, Object> worlds, Object type, Object world) {
-		worlds.put(type, world);
-		WorldEvents.onWorldLoad((ServerWorld) world);
-
-		return world;
+	// Fabric usually fires the event much earlier in the method, so this is just picking a point closer to when Forge would fire it.
+	@Inject(method = "createWorlds", at = @At(value = "INVOKE", target = "net/minecraft/world/dimension/DimensionType.getAll ()Ljava/lang/Iterable;"))
```
i509VCB, line 52: Shouldn't the inject target be in `Lnet/minecraft/whatever` format?

kitlith: hm. good question, and i hope i haven't goofed my previous PRs because of this. I've just been copying from IDEA's bytecode viewer: `INVOKESTATIC net/minecraft/world/dimension/DimensionType.getAll ()Ljava/lang/Iterable;`

kitlith: Well, it at least seems to work as intended as-is. I'll leave "should it be changed anyway" as something for future consideration.

TheGlitch76: i con't care about the signature format as long as it's readable and it applies

PR #151, opened 2020-07-25
Use raw types to "hand erase" ExtendedForgeRegistryEntry's generics

famous1622:
Fixes the "TNT Yeeter" edge case that #141 can't replicate correctly. Changes the ExtendedForgeRegistryEntry generic from `<V>` to `<V extends IForgeRegistryEntry<V>>`, and converts all the Mixins to use raw types. I wrote a test mod and ran it over this and it was fine, but given how fragile replicating the ABI here has turned out to be, the more test cases the better.


TheGlitch76:
watch as this breaks something else

PR #154, opened 2020-07-27
Add support for keybindings

Snakefangox:
This PR adds a new module to allow patchwork to support keybindings, this will prevent mods that use forge's special handling from crashing however it does not completely implement this special handling. The two additions in question are conflict handling and modifier key support. Because forge uses the default KeyBinding methods for this behavior this shouldn't cause any issues. I can make a follow up PR to add the complete API but these methods are not commonly used to begin with and I suspect this might be all that is needed. Additionally this method allows fabric mods that add these functions to support Patchwork mods.

I'm not sure about Patchwork naming conventions so better names are welcome.

This requires a Patcher PR, namely https://github.com/PatchworkMC/patchwork-patcher/pull/67.


patchwork-key-bindings/build.gradle
```diff
@@ -0,0 +1,2 @@
+archivesBaseName = "patchwork-gui"
```
Rongmario, line 1: `patchwork-key-bindings`

Snakefangox: Ah

patchwork-key-bindings/src/main/java/net/patchworkmc/mixin/keybindings/MixinKeybinding.java
```diff
@@ -0,0 +1,88 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import net.minecraft.client.options.KeyBinding;
+import net.minecraft.client.util.InputUtil;
+
+import net.patchworkmc.impl.keybindings.ForgeKeyBindingConstruct;
+
+@Mixin(KeyBinding.class)
+public abstract class MixinKeybinding implements Comparable<KeyBinding>, IForgeKeybinding, ForgeKeyBindingConstruct {
```
famous1622, line 35: Is there any reason this doesn't match the casing of KeyBinding?

patchwork-key-bindings/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,27 @@
...
+  "contact": {
+    "issues": "https://github.com/PatchworkMC/patchwork-api/issues",
+    "sources": "https://github.com/PatchworkMC/patchwork-api"
+  },
+  "authors": [
+    "PatchworkMC"
+  ],
+  "depends": {
+    "fabricloader": ">=0.8.4",
+    "patchwork-fml": "*"
```
famous1622, line 17: You don't depend on patchwork-fml

patchwork-key-bindings/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,27 @@
+{
+  "schemaVersion": 1,
+  "id": "patchwork-key-bindings",
+  "name": "Patchwork Keybindings",
+  "version": "${version}",
+  "license": "LGPL-2.1-only",
+  "icon": "assets/patchwork-gui/icon.png",
```
famous1622, line 7: Wrong folder

patchwork-key-bindings/src/main/java/net/patchworkmc/mixin/keybindings/MixinKeybinding.java
```diff
@@ -0,0 +1,88 @@
...
+	@Shadow
+	private InputUtil.KeyCode keyCode;
+
+	//These exist in forge to allow modifiers and conflicting keys to work, this is not implemented
+	//but these remain to avoid stubbing the methods or breaking mods that use these directly for some reason
+	private KeyModifier keyModifierDefault = KeyModifier.NONE;
+	private KeyModifier keyModifier = KeyModifier.NONE;
+	private IKeyConflictContext keyConflictContext = KeyConflictContext.UNIVERSAL;
```
Rongmario, line 43: `@Unique` would be good

patchwork-key-bindings/src/main/resources/fabric.mod.json
Rongmario, line 7: `patchwork-gui` not fixed here

patchwork-key-bindings/src/main/java/net/patchworkmc/mixin/keybindings/MixinKeybinding.java
Rongmario: Also see `IForgeKeybinding`, can't fix that, so: either pick `Keybinding` or `KeyBinding`

florensie: I'd say MixinKeyBinding here since the general rule for mixins is `MixinClassName`.
```suggestion
public abstract class MixinKeyBinding implements Comparable<KeyBinding>, IForgeKeybinding, ForgeKeyBindingConstruct {
```

patchwork-key-bindings/src/main/java/net/patchworkmc/impl/keybindings/ForgeKeyBindingConstruct.java
```diff
@@ -0,0 +1,27 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+package net.patchworkmc.impl.keybindings;
+
+import net.minecraftforge.client.settings.IKeyConflictContext;
+import net.minecraftforge.client.settings.KeyModifier;
+
+public interface ForgeKeyBindingConstruct {
+	void constructForgeKeyBindingOptions(IKeyConflictContext keyConflictContext, KeyModifier keyModifier);
```
rikka0w0, line 26: I would like to suggest that we should add a "patchwork$" prefix, if this method is not a part of Forge. But I think this name is okay.

famous1622: I don't feel like it's necessary here, not too sure what the rules in this are though

TheGlitch76: prefix if it's not from forge

Snakefangox:
Have added _some_ of the functionality. Namely that forge allows you to categorize key binds so they only fire in certain situations and not respecting that may have some weird side effects like jetpack mods working with your inventory open. Don't think it's too big a deal but it's a tiny change, sorry if this is bad form.

patchwork-key-bindings/src/main/java/net/patchworkmc/mixin/keybindings/MixinKeyBinding.java
```diff
@@ -0,0 +1,103 @@
...
+
+import net.patchworkmc.impl.keybindings.ForgeKeyBindingConstruct;
+
+@Mixin(KeyBinding.class)
+public abstract class MixinKeyBinding implements Comparable<KeyBinding>, IForgeKeybinding, ForgeKeyBindingConstruct {
+	@Shadow
+	private InputUtil.KeyCode keyCode;
+
+	//These exist in forge to allow modifiers and conflicting keys to work, this is not implemented
+	//but these remain to avoid stubbing the methods or breaking mods that use these directly for some reason
```
TheGlitch76, line 44: @Unique breaks reflection, which is fine. We can remove it if a mod reflects here for some reason.
```suggestion
	// These exist in forge to allow modifiers and conflicting keys to work, this is not implemented
	// but these remain to avoid stubbing the methods
```

PR #156, opened 2020-07-28
AnimalTameEvent 2: Electric Boogaloo

famous1622:
Same thing as last time, just replacing the killer Inject from MixinTameableEntitySubclass with a redirect that doesn't crash the game on load. You might want to test this to be super duper sure but it should be fine this time.


TheGlitch76:
Ran out of time to test this tonight.

PR #157, opened 2020-07-28
Create patchwork-api-base

TheGlitch76:
This module does two things:
- Keeps all our outside dependency declarations in one place
- Provides no-overhead classes Forge mods and Patchwork modules might need (i.e. constants, EnvType, etc. that doesn't really belong in the FML module)

Currently only the first is done to keep noise down.
Note git got confused when I was moving some icons that were not in the right folders.


kitlith:
Hm. I have mixed feelings about this.

On the one hand, I understand why you want to do something like this. On the other hand, it might be nice to eventually specify which modules of fabric api each of our patchwork modules depends on.

But hey, we can always change it again later, and I guess it's not like patchwork's modules are actually meant to be used independently anyway.

TheGlitch76:
"If we include the whole API we're going to use the whole API"
in other words: Until some actually decent solution of using individual modules exists, we may as well act like every other fabric mod here.

rikka0w0:
Should we put some commonly used code here, like the `public static final String` for method/field signatures?

TheGlitch76:
method/field signatures tend to be specific to one module, it's easier to keep them there. Also avoids merge conflicts (see god classes)

patchwork-capabilities/build.gradle
```diff
@@ -2,5 +2,6 @@ archivesBaseName = "patchwork-capabilities"
 version = getSubprojectVersion(project, "0.2.0")
 
 dependencies {
-    compile project(path: ':patchwork-fml', configuration: 'dev')
+	implementation project(path: ':patchwork-api-base', configuration: 'dev')
```
Rongmario, line 5: Mixed tabs + spaces

kitlith:
^ if we're going to make everything depend on base anyway, then having them all share the single icon might be nice.

TheGlitch76:
forgot to put it here: sharing icons isn't currently possible

PR #159, opened 2020-07-28
Fixed #114

### LINKED ISSUE #114: Forge distinguishes between "onLivingAttack" and "onPlayerAttack" while patchwork doesn't.
YarnForge's ForgeHooks:
```java
        public static boolean onLivingAttack(LivingEntity entity, DamageSource src, float amount) {
                return entity instanceof PlayerEntity || !MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, amount));
        }

        public static boolean onPlayerAttack(LivingEntity entity, DamageSource src, float amount) {
                return !MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, amount));
        }
```

Patchwork's EntityEvents:
```java
	public static boolean onLivingAttack(LivingEntity entity, DamageSource src, float damage) {
		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
	}
```

The LivingEntity mixin ends up not checking if the entity is `instanceof PlayerEntity` while forge does. For all I know there's a reason for this, but it's not documented if there is a reason.

obj-obj:
Has no effect on vanilla. I haven't tested with mods yet (because the creator of the issue hasn't given an example of a mod that needs this event), but the code is the same as yarnforge now, which means that it's probably fixed.

Intellij also optimized the imports, which is good I guess.

### DIFF
--- a/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
+++ b/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
@@ -107,6 +107,10 @@ public class EntityEvents implements ModInitializer {
 		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
 	}
 
+	public static boolean onPlayerAttack(LivingEntity entity, DamageSource src, float amount) {
+		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, amount));
+	}
+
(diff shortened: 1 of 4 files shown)


obj-obj:
Also, this is my first contribution :)

Rongmario:
Yeah, checkstyle didn't pass because of the starred import :p

obj-obj:
So does it really matter if checkstyle fails because of star imports?

Or should I revert the imports and make a new PR

obj-obj:
noooooo

obj-obj:
ok I broke it
gotta refork

PR #160, opened 2020-07-28
Fixed #114 (works in vanilla, not tested with mods) (v2)

### LINKED ISSUE #114: Forge distinguishes between "onLivingAttack" and "onPlayerAttack" while patchwork doesn't.

obj-obj:
This is my first contribution :)

### DIFF
--- a/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
+++ b/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
@@ -107,8 +107,8 @@ public class EntityEvents implements ModInitializer {
 		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
 	}
 
-	public static boolean onPlayerAttack(LivingEntity entity, DamageSource src, float amount) {
-		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, amount));
+	public static boolean onPlayerAttack(LivingEntity entity, DamageSource src, float damage) {
+		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
 	}
 
 	public static void onLivingSetAttackTarget(LivingEntity entity, LivingEntity target) {
--- a/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
+++ b/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
@@ -104,11 +104,11 @@ public class EntityEvents implements ModInitializer {
 	}
 
 	public static boolean onLivingAttack(LivingEntity entity, DamageSource src, float damage) {
-		return entity instanceof PlayerEntity || MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
+		return entity instanceof PlayerEntity || !MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
 	}
 
 	public static boolean onPlayerAttack(LivingEntity entity, DamageSource src, float damage) {
-		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
+		return !MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
 	}
 
 	public static void onLivingSetAttackTarget(LivingEntity entity, LivingEntity target) {
--- a/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
+++ b/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
@@ -104,7 +104,7 @@ public class EntityEvents implements ModInitializer {
 	}
 
 	public static boolean onLivingAttack(LivingEntity entity, DamageSource src, float damage) {
-		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
+		return entity instanceof PlayerEntity || MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
 	}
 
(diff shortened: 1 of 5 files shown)


patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
```diff
@@ -107,6 +107,10 @@ public static boolean onLivingAttack(LivingEntity entity, DamageSource src, floa
 		return MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage));
```
TheGlitch76, line 107: This needs to be changed to short-circuit like in the issue

patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
```diff
@@ -107,6 +107,10 @@ public static boolean onLivingAttack(LivingEntity entity, DamageSource src, floa
...
 	}
 
+	public static boolean onPlayerAttack(LivingEntity entity, DamageSource src, float damage) {
```
TheGlitch76, line 110: This needs a entry in the patchwork-god-classes module

obj-obj: I do have it in ForgeHooks, is that where it's supposed to be, or should it be somewhere else?

obj-obj:
Ok I broke entity attacks (entites cannot attack at all)
Gotta fix that

patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
kitlith: They changed the relevant method in ForgeHooks (I implemented both originally.) However, you should change the other method in ForgeHooks to call the correct method in EntityEvents, (and remove the now redundant check from there)

obj-obj: Should I invert the result like onPlayerAttack or not?

kitlith:
Please note that the original ForgeHooks returns the inverse of what the method on EntityEvents returns, which is almost certainly why this is breaking as you didn't account for it.

patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
kitlith: The overall result should be the same as if you called it on Forge.

obj-obj: The problem is, if I implement it the same as forge, it prevents all entities in the world from attacking...

kitlith: You're not implementing it the same as forge though.

Forge is:
```java
entity instanceof PlayerEntity || !MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage))
```

while you currently effectively have:
```java
!(entity instanceof PlayerEntity || MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage)))
```

please note the location of the inversions.

obj-obj: I've already tested doing the same thing as forge locally, it just prevents me from attacking and entities from attacking

kitlith: Then please at least commit that behavior, because it is currently definately incorrect.

obj-obj: I've updated it, although neither the player nor entities can attack now.

obj-obj:
for SOME REASON
when a specific line of code is `MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage))`, players can't attack
But when it's `!(!MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage)))`, players can attack

patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
kitlith: Yes, that's because given how this method is used, the logic is now `!(entity instanceof PlayerEntity || !MinecraftForge.EVENT_BUS.post(new LivingAttackEvent(entity, src, damage)))`. So, either change how the method is used (remove the inversion) or invert the entire result.

obj-obj:
I need to actually learn how forge works before I try to fix bugs lmao

kitlith:
For the record, @obj-obj, afaict all you needed was as extra inversion there.

Patchwork's version of that hook doesn't seem to follow the same convention as Forge's version (specifically, one returns the inverse of the other). That might be worth fixing too (at it's callsites) but as I don't think I was the one to put those hooks in place, I can't speak for why they don't follow forge.

PR #161, opened 2020-07-29
Add classes and fields to ForgeHooks, which might be used by hacky mods

rikka0w0:
This fix makes https://github.com/MinecraftModDevelopmentMods/AdditionalLootTables/tree/master-1.15.2 work.
By design, AdditionalLootTables 1.15.2 should also work in 1.14.4. Without coremoding, in order to append items to an existing loottable, Forge mods have to use reflection on `ForgeHooks.lootContext`.


kitlith:
Don't have to deal with this right now, but I wonder how #139 would change things, since that isn't using the LootTableContext struct anymore internally.

Overall, the changes here make sense to me.

PR #162, opened 2020-07-29
Feature/extensions blockentity

rikka0w0:
This module implements:

1. Everything in `IForgeTileEntity` except those `IModelData` stuff.
2. Hooks for BlockEntityRenderer registration. (We cannot reuse Fabric hooks for this, the Fabric registration event is fired much earlier than Forge's)
3. FastTESR in 1.14.4, will be removed and replaced with vanilla BlockEntityRenderer in 1.15 and above.

This pr is not huge, the licence header contributes a lot to the number of addition....


patchwork-extensions-blockentity/build.gradle
```diff
@@ -0,0 +1,6 @@
+archivesBaseName = "patchwork-extensions-blockentity"
+version = getSubprojectVersion(project, "0.1.0")
+
+dependencies {
+	compile project(path: ':patchwork-capabilities', configuration: 'dev')
```
rikka0w0, line 5: This dependency is needed because:
```
public interface IForgeTileEntity extends ICapabilitySerializable<CompoundTag>
```
However, it might be possible to remove it by moving some of the default impls from IForgeTileEntity to the capability package. I would like to hear other's suggestion on this.

TheGlitch76: This is fine how it is IMO, if you're going to need IForgeTileEntity you're going to need capabilities. Not adding this dependency would just make things confusing

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	/**
+	 * Return an {@link AxisAlignedBB} that controls the visible scope of a {@link TileEntitySpecialRenderer} associated with this {@link TileEntity}
+	 * Defaults to the collision bounding box {@link Block#getCollisionBoundingBoxFromPool(World, int, int, int)} associated with the block
+	 * at this location.
+	 *
+	 * @return an appropriately size {@link AxisAlignedBB} for the {@link TileEntity}
+	 */
+	// @OnlyIn(Dist.CLIENT)
```
TheGlitch76, line 110: why comment this out?

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
```diff
@@ -0,0 +1,197 @@
...
+			bb = INFINITE_EXTENT_AABB;
+		} else if (block != null && block != Blocks.BEACON) {
+			Box cbb = null;
+
+			try {
+				cbb = state.getCollisionShape(getTileEntity().getWorld(), pos).getBoundingBox().offset(pos);
+			} catch (Exception e) {
+				// We have to capture any exceptions that may occur here because BUKKIT servers like to send
+				// the tile entity data BEFORE the chunk data, you know, the OPPOSITE of what vanilla does!
+				// So we can not GARENTEE that the world state is the real state for the block...
```
TheGlitch76, line 131:
```suggestion
				// So we can not GUARANTEE that the world state is the real state for the block...
```

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
```diff
@@ -0,0 +1,197 @@
...
+				// So, once again in the long line of US having to accommodate BUKKIT breaking things,
+				// here it is, assume that the TE is only 1 cubic block. Problem with this is that it may
+				// cause the TileEntity renderer to error further down the line! But alas, nothing we can do.
+				cbb = new net.minecraft.util.math.Box(pos.add(-1, 0, -1), pos.add(1, 1, 1));
```
TheGlitch76, line 135: we're not writing patches

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinClientPlayNetworkHandler.java
```diff
@@ -0,0 +1,121 @@
...
+import net.minecraft.client.world.ClientWorld;
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.packet.s2c.play.BlockEntityUpdateS2CPacket;
+import net.minecraft.network.packet.s2c.play.ChunkDataS2CPacket;
+import net.minecraft.util.math.BlockPos;
+
+import net.fabricmc.fabric.api.block.entity.BlockEntityClientSerializable;
+
+/**
+ * TODO: This Mixin can be less hacky if we can talk to the Fabric API team.
```
TheGlitch76, line 45: What needs to be done here?
If this is being done to work around a FAPI patch, could mixin priorities be used to apply over their patches?

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinClientPlayNetworkHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@Redirect(method = "onChunkData", at = @At(value = "INVOKE", target = ClientWorld_getBlockEntity))
+	private BlockEntity onChunkData_world_getBlockEntity(ClientWorld clientWorld, BlockPos blockPos) {
+		CompoundTag blockEntityTag = onChunkData_BETag.get();
+		onChunkData_BETag.remove();
+
+		ClientPlayNetworkHandler me = (ClientPlayNetworkHandler) (Object) this;
+		BlockEntity blockEntity = me.getWorld().getBlockEntity(blockPos);
+
+		if (blockEntity != null && !(blockEntity instanceof BlockEntityClientSerializable)) {
```
TheGlitch76, line 79: This looks like a hack. Would a marker interface in Patcher be helpful?

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinWorld.java
```diff
@@ -0,0 +1,178 @@
...
+	/////////////////////////////////////
+	/// addBlockEntity()
+	/////////////////////////////////////
+	@Inject(method = "addBlockEntity", at = @At("HEAD"))
+	private void onAddBlockEntity(BlockEntity blockEntity, CallbackInfoReturnable<Boolean> ci) {
+		World me = (World) (Object) this;
+
+		if (blockEntity.getWorld() != me) {
+			// Forge - set the world early as vanilla doesn't set it until next tick
+			blockEntity.setWorld(me);
```
TheGlitch76, line 61: could this cause incompatibilities in fabric mods?

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
famous1622: Fabric's @Environment?

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinClientPlayNetworkHandler.java
rikka0w0: We need to insert our own hook here. We cannot reuse Fabric's BE sync, because Fabric API requires BE to be sync to implement BlockEntityClientSerializable. We cannot make IForgeTileEntity extends BlockEntityClientSerializable, otherwise every single BlockEntity will implement BlockEntityClientSerializable....

Ideally we should let Fabric's hook checks for BlockEntityClientSerializable, if the BE is not a BlockEntityClientSerializable, then it should call our hook. This part of logic is in Fabric API's mixin, so we cannot mixin it. What I would expect is that, Fabric API moves that logic from the mixin class to a separate normal class, so that we can mixin it. 

Actually everything is working here and we don't need to set priority at all. I chose an Injection point so that our hook will be executed earlier than Fabric API's. We will then check if the BE implements BlockEntityClientSerializable, if that's the case, pass it to Fabric API, otherwise, we handle it and pass it to IForgeTileEntity. Vanilla BEs are handled in onChunkData, so they don't need this kind of callback.

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinClientPlayNetworkHandler.java
rikka0w0: explained in the previous comment

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinWorld.java
rikka0w0: No, the world field will be set anyway. Fabric mods are made under the assumption that world is not available upon creation, so they will not attempt to access it that early. I think this modification should not cause trouble to fabric mods. But I'm sure that a lot of Forge mods need this.

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
rikka0w0: does the patcher transform @OnlyIn to @Environment?

TheGlitch76: yes

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
rikka0w0: =This does not change any vanilla rendering behavior. I would suggest we should keep this. Or simply set cbb to null, so that when an exception is captured, the BE renderer will always be invoked.

TheGlitch76:
Please update to use patchwork-api-base

rikka0w0:
Updated for patchwork-api-base

patchwork-extensions-blockentity/src/main/java/net/minecraftforge/common/extensions/IForgeTileEntity.java
TheGlitch76: i meant import instead of `net.minecraft.util.math`

rikka0w0: fixed

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinWorld.java
```diff
@@ -0,0 +1,178 @@
...
+	@Inject(method = "tickBlockEntities", at = @At(value = "INVOKE", ordinal = 0, shift = Shift.BEFORE,
+			target = "java/util/List.removeAll(Ljava/util/Collection;)Z"))
+	private void onBlockEntitiesRemoved(CallbackInfo ci) {
+		for (BlockEntity blockEntity: unloadedBlockEntities) {
+			((IForgeTileEntity) blockEntity).onChunkUnloaded();
+		}
+	}
+
+	@Unique
+	private static final ThreadLocal<BlockEntity> onBlockEntitiesRemoved_BlockEntity = ThreadLocal.withInitial(() -> null);
```
TheGlitch76, line 110: I don't think this part of World is thread-safe. Maybe implement something like what I do in my BlockSnapshot PR?

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinWorldRenderer.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	/////////////////////////////////////
+	/// renderPartiallyBrokenBlocks()
+	/////////////////////////////////////
+	@Shadow
+	private ClientWorld world;
+	@Unique
+	private static final ThreadLocal<BlockPos> blockPosParam = ThreadLocal.withInitial(() -> null);
```
TheGlitch76, line 68: again, i'm 90% sure this is not thread safe

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinWorld.java
rikka0w0: `tickBlockEntities` is supposed to be fired on either the main server thread or the main client thread?

image

patchwork-extensions-blockentity/src/main/java/net/patchworkmc/mixin/extensions/blockentity/MixinWorldRenderer.java
rikka0w0: here I only collect the value of a local variable, there's no access to the internal state of other objects, I don't get why this is unsafe.

TheGlitch76:
not going to worry about unneeded threadlocals for now.
They're not that hard to remove if a profiler says they're an issue

TheGlitch76:
Merge conflicts

PR #163, opened 2020-07-30
Organized ForgeHooks

obj-obj:


obj-obj:
Why does it show my old username on the commits...

kitlith:
Your local git configuration might have the wrong name/email setup, as that's what github uses to detech what commits came from who.

obj-obj:
Ah, it did still have my old user
fixed that

obj-obj:
oops

obj-obj:
This isn't going to be approved so im just gonna close it
Doesn't really have a point anyway

PR #174, opened 2020-08-09
Feature/block harvest

rikka0w0:
Remove the malfunction 1.15 patch,

Move the BlockEvent from events-world to extensions-block, this can remove an unnecessary dependency, to make porting easier.


patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/harvest/MixinClientPlayerInteractionManager.java
```diff
@@ -0,0 +1,57 @@
...
+import net.minecraft.client.MinecraftClient;
+import net.minecraft.client.network.ClientPlayerInteractionManager;
+import net.minecraft.entity.player.PlayerEntity;
+import net.minecraft.fluid.FluidState;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.World;
+
+import net.patchworkmc.impl.extensions.block.Signatures;
+
+// This is a 1.15+ patch, should not be used in 1.14.4
```
TheGlitch76, line 40: Go ahead and just remove this.

patchwork-extensions-block/src/main/java/net/patchworkmc/mixin/extensions/block/harvest/MixinServerPlayerInteractionManager.java
```diff
@@ -0,0 +1,151 @@
...
+	private boolean patchwork$tryBreakBlock_removeBlock(ServerWorld world, BlockPos pos, boolean bool) {
+		return true; // bypass if (bl && bl2) {
+	}
+
+	@Redirect(method = "tryBreakBlock", at = @At(value = "INVOKE", target = Signatures.Block_onBroken, ordinal = 0))
+	private void patchwork$tryBreakBlock_onBroken(Block block, IWorld world, BlockPos pos, BlockState state) {
+		// Suppress this call
+	}
+
+	@Redirect(method = "tryBreakBlock", at = @At(value = "INVOKE", target = Signatures.ServerPlayerInteractionManager_isCreative, ordinal = 0))
```
TheGlitch76, line 105: I don't think this needs to be a redirect. One option could be to @Slice from the isCreative check and inject at RETURN ordinal = 0.

PR #175, opened 2020-08-11
Make CapabilityRegisteredCallback store class names not class objects

famous1622:
This is needed so the CapabilityInject handling in Patcher doesn't have to load types onto the stack, and replicate Forge behavior of not needing a CapabilityInject's type to exist at runtime.

### DIFF
--- a/patchwork-capabilities/src/main/java/net/patchworkmc/api/capability/CapabilityRegisteredCallback.java
+++ b/patchwork-capabilities/src/main/java/net/patchworkmc/api/capability/CapabilityRegisteredCallback.java
@@ -29,6 +29,10 @@ public interface CapabilityRegisteredCallback<C> {
 	void onCapabilityRegistered(Capability<C> capability);
 
 	static <C> Event<CapabilityRegisteredCallback<C>> event(Class<C> type) {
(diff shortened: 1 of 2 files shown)


patchwork-capabilities/src/main/java/net/patchworkmc/api/capability/CapabilityRegisteredCallback.java
```diff
@@ -29,6 +29,10 @@
 	void onCapabilityRegistered(Capability<C> capability);
 
 	static <C> Event<CapabilityRegisteredCallback<C>> event(Class<C> type) {
-		return CapabilityRegisteredCallbackInternal.getOrCreateEvent(type);
+		return CapabilityRegisteredCallbackInternal.getOrCreateEvent(type.getName());
+	}
+
+	static <C> Event<CapabilityRegisteredCallback<C>> event(String className) {
+		return CapabilityRegisteredCallbackInternal.getOrCreateEvent(className);
```
TheGlitch76, line 36: I'm not confident we should allow arbitrary strings, but i guess in the long run it's really not that big of a deal

PR #177, opened 2020-08-13
Implement ClientPlayerNetworkEvent

WalkerKnapp:
Implements ClientPlayerNetworkEvent in patchwork-networking.

I am still unsure about if this is the best package to use for this event, so if it would be better to use another mod or to create a new one, just let me know.

Corresponding testmods PR: PatchworkMC/patchwork-testmods#4


rikka0w0:
I would like to suggest that you should move ClientHooks from then networking module to the god-classes module:
https://github.com/PatchworkMC/patchwork-api/tree/master/patchwork-god-classes/src/main/java/net/minecraftforge/fml/hooks

WalkerKnapp:
Done 👍

Issue #94, opened 2020-06-21
Required events for Forestry


ramidzkh:
[Forestry](https://github.com/ForestryMC/ForestryMC/tree/dev-1.15)

These are only the events which are remaining as of now

- [ ] `BlockEvent.HarvestDropsEvent`
- [ ] `ChunkDataEvent.Load`
- [ ] `ChunkDataEvent.Save`
- [ ] `ChunkEvent.Load`
- [ ] `EntityItemPickupEvent`
- [ ] `EntityTravelToDimensionEvent`
- [ ] `LootTableLoadEvent`
- [ ] `ModelBakeEvent`
- [ ] `ModelRegistryEvent`
- [ ] `PlayerEvent.PlayerChangedDimensionEvent`
- [ ] `RenderGameOverlayEvent.Post`
- [ ] `RenderWorldLastEvent`
- [ ] `TickEvent.RenderTickEvent`

ZNixian:
You've linked to the 1.15 dev branch?

ramidzkh:
Sssh they're identical

rikka0w0:
These should be implemented in a new package (ModelLoader):
ModelBakeEvent
ModelRegistryEvent

famous1622:
I can get RenderGameOverlayEvent if no one else is, have a decent solution to forge straight up reimplementing the class

valoeghese:
I started implementing chunk data events the other day

PR #179, opened 2020-08-17
Add EntityItemPickupEvent and EntityTravelToDimensionEvent

williambl:
Both needed for Forestry (#94)

### DIFF
--- a/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
+++ b/patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
@@ -61,6 +61,7 @@ import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.entity.projectile.ExplosiveProjectileEntity;
 import net.minecraft.entity.projectile.ProjectileEntity;
 import net.minecraft.entity.thrown.ThrownEntity;
+import net.minecraft.entity.vehicle.StorageMinecartEntity;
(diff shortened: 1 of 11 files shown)


williambl:
Okay, this is ready now.

patchwork-events-entity/src/main/java/net/patchworkmc/impl/event/entity/EntityEvents.java
```diff
@@ -217,6 +219,19 @@ public static boolean onProjectileImpact(ThrownEntity throwable, HitResult ray)
 		return MinecraftForge.EVENT_BUS.post(new ProjectileImpactEvent.Throwable(throwable, ray));
 	}
 
+	public static boolean onTravelToDimension(Entity entity, DimensionType dimensionType) {
+		EntityTravelToDimensionEvent event = new EntityTravelToDimensionEvent(entity, dimensionType);
+		return !MinecraftForge.EVENT_BUS.post(event);
+		/* Forge has this thing:
```
TheGlitch76, line 225: why comment this out?

williambl: `StorageMinecart` doesn't seem to have `dropContentsWhenDead` - i assume it's a forge method

williambl: It's the same behaviour as forge now - what do you mean by adding a // TODO? What should it say?

PR #186, opened 2020-09-04
Fix energy capability not being registered and mark `@CapabilityInject` as deprecated

williambl:


famous1622:
LGTM, we should probably get rid of that annotation though (unless some mod looks it up at runtime for some reason)

TheGlitch76:
oops, the github actions don't exclude checkstyle for the building.

Could you mark CapabilityInject as @Deprecated while you're fixing checkstyle?

williambl:
Okay.

TheGlitch76 (approved):

Issue #125, opened 2020-07-12
Required for WorldEdit


ramidzkh:
This is for the Forge edition of WorldEdit

- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/patches/minecraft/net/minecraft/client/gui/widget/AbstractButtonWidget.java.patch#L21>
- [ ] Publicize `MinecraftClient#textureManager`
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java#L337>
- [ ] Fix `CapabilityProviderHolder`
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/fmllauncher/java/net/minecraftforge/fml/loading/FMLLoader.java>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/patches/minecraft/net/minecraft/world/World.java.patch#L94>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/patches/minecraft/net/minecraft/world/World.java.patch#L52>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/main/java/net/minecraftforge/common/util/FakePlayer.java>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/patches/minecraft/net/minecraft/world/TestableWorld.java.patch#L8>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/patches/minecraft/net/minecraft/server/world/ServerWorld.java.patch#L329>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/main/java/net/minecraftforge/fml/event/server/FMLServerStoppingEvent.java>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/main/java/net/minecraftforge/event/entity/player/PlayerInteractEvent.java#L292>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/main/java/net/minecraftforge/event/CommandEvent.java>
- [ ] <https://github.com/PatchworkMC/YarnForge/blob/1.14.x/src/main/java/net/minecraftforge/fml/ModContainer.java#L118>

VikoRifeo:
[...]

PR #190, opened 2020-09-15
Publicized MinecraftClient#textureManager

YTG1234:
Just saw this on #125 so I decided to do it.


YTG1234:
Maybe we could also move all Forge's access modifications that are already in Patchwork to this new module, idk how it's going to be possible to "load" the access widener when developing other modules. Any thoughts?

patchwork-access-modifications/build.gradle
```diff
@@ -0,0 +1,10 @@
+archivesBaseName = "patchwork-access-modifications"
+version = getSubprojectVersion(project, "0.0.0")
```
famous1622, line 2: The version should start at 0.1.0

settings.gradle
```diff
@@ -45,3 +45,4 @@ include 'patchwork-resource'
 include 'patchwork-tags'
 include 'patchwork-tooltype'
 include 'patchwork-vanilla-patches'
+include 'patchwork-access-modifications'
```
famous1622, line 48: These should be in alphabetical order.

famous1622 (changes requested):

settings.gradle
YTG1234: Should be fixed now

patchwork-access-modifications/build.gradle
YTG1234: Should be fixed now

TheGlitch76:
Moving access wideners to their own module isn't the right solution yet because nobody has made a jar processor to make access wideners transitive.
I think instead of its own module, for now this should either be in an appropriate module that already exists (Check the Forge source code and see if it's used anywhere), or failing that dropped in `patchwork-fml`.

YTG1234:
I think it may be used in the forge `CloudRenderer`: `mc.textureManager.bindTexture(texture);`. I'm not sure

YTG1234:
Yep. It's only used in `CloudRenderer`

YTG1234:
I will put it in the FML module

YTG1234:
Moved it to the FML module for now. I didn't bump the version - as the commit msg says.

YTG1234:
Turns out I put a `/`. Fixed it to a `.`, it should run now.

YTG1234:
Updated the title so it's less confusing

YTG1234:
@famous1622 I changed things up a little, can you review again?

rikka0w0:
> Moving access wideners to their own module isn't the right solution yet because nobody has made a jar processor to make access wideners transitive.
> I think instead of its own module, for now this should either be in an appropriate module that already exists (Check the Forge source code and see if it's used anywhere), or failing that dropped in `patchwork-fml`.

I think in the future the rendering stuff should have their own package. But for now I would agree that this aw should be in `patchwork-fml`. Sometimes `mc.textureManager` is used by mods in their tileentity render and entity render, I don't there is a direct reference to that field in Forge.

YTG1234:
It's interesting that in most places Forge uses `getTextureManager()`, so I don't understand why the field has been made public

famous1622:
Was probably public forever ago and they don't want to break mods

YTG1234:
Test

TheGlitch76 (approved):
LGTM!

YTG1234:
> Was probably public forever ago and they don't want to break mods

I thought Forge was the opposite of backwards-compatibility

TheGlitch76:
Forge is more than just Lex and all of them do try to maintain back-compat for some things.

PR #193, opened 2020-10-04
Fix item pickup events

famous1622:
EntityItemPickupEvent and PlayerEvent.ItemPickupEvent were implemented incorrectly, leading to no items being able to be picked up and events firing at times they shouldn't. This PR fixes the implementation of these two events.

### DIFF
--- a/patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
+++ b/patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
@@ -20,6 +20,8 @@
 package net.patchworkmc.mixin.event.entity;
 
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.Unique;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.At.Shift;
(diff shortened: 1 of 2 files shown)


patchwork-events-entity/src/main/java/net/patchworkmc/mixin/event/entity/MixinItemEntity.java
```diff
@@ -35,34 +37,73 @@
...
+	@Inject(method = "onPlayerCollision", at = @At("HEAD"), cancellable = true)
+	private void patchwork_checkForPickupDelay(PlayerEntity player, CallbackInfo ci) {
+		hasPickupDelay = this.pickupDelay > 0;
+	}
 
 	@Inject(method = "onPlayerCollision",
-			at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;getCount()I"),
-			cancellable = true
+					at = @At(value = "INVOKE", target = "Lnet/minecraft/item/ItemStack;getCount()I"),
+					cancellable = true
```
ZNixian, line 69: These lines are being re-indented - any particular reason?

TheGlitch76 (approved):
Sorry for taking so long! The indention is a little funky on this but it's a pretty bad bug so i'll leave it to be fixed Later:tm:
Please bother me on Discord about PR reviews if I don't even remark that I'm waiting on something, it's very likely I've just forgotten about it

PR #201, opened 2020-11-14
Allow for subprojects that aren't fabric mods

kitlith:
This is accomplished by moving away from subprojects and allprojects closures to using precompiled script plugins to accomplish the same thing. We can now opt out of the things in subprojects/allprojects by simply not applying the relevant plugin.

loom shareCaches won't quite work (and thus running this build in parallel) until https://github.com/FabricMC/fabric-loom/pull/299 is merged and present in a release of fabric-loom.

(Motivated by wanting to create an annotation processor as a subproject, which is decidedly *not* a fabric mod.)

PR #202, opened 2020-11-16
God class codegen

kitlith:
This is basically my first working version of stuff. depends on #201 , as it uses an annotationProcessor defined in a subproject and makes use of the precompiled script plugins to make it easier to e.g. disable checkstyle for a particular module.

#### Where to go from here: (ideas listed in order of most to least tempting)
 - We should probably throw away generation of the json files (and thus the annotation processor?), i think it's an intermediate step that is simply complicating things.
   - Alternatively, the annotation processor generates a set of class files per module, which are later merged into a single set of class files in god-classes, see generating bytecode below
 - switch source generation libraries? I'm currently using jcodemodel from https://github.com/eclipse-ee4j/jaxb-ri, but javapoet may be better/more popular and may address some of the issues i was running into.
 - Glitch suggested generating bytecode instead of source files (which I think may smooth some things out)
   - NuclearFarts claims that this would mesh well with IDEs. (personally, i don't see why it would work any worse than source generation but hell if i know)
 - VFMADDSUBPD brought up the possibility of just using Mixins for all of this, but I think we all agree that codegen is more graceful.
 - I brought up the possibility of going in the middle, and basically having some sort of "code fragment" system -- we still write all the parts of the god-classes manually, we just divide it up between all of the modules and bring it together later.


kitlith:
Before I forget, I'm gonna change the annotation next time I work on this so you always include the name of the item you're adding, to make it easier to search for the items.

basically the biggest upside to this method is also its biggest downside, in that everything is spread everywhere.

kitlith:
Closing in favor of #203. I wouldn't mind revisiting this again another time, but at the moment I feel the biggest upside of this is also it's biggest disadvantage, and the other method is less likely to bite us in the ass later. i think. i hope.

kitlith:
reopening i guess since at least @famous1622 is still interested in this.

PR #203, opened 2020-11-16
Completely stub all classes in god-classes

kitlith:
~~NOTE: current commit only stubs ForgeHooksClient, just to get an idea of what this would look like.~~

Alternative to #202, basically a dead simple approach to reducing conflicts by adding more context.

Somewhat tedious to setup, but there's no underlying complexity that could bite us in the ass later.

Also introduces an annotation to mark stubbed items so that we can detect them in other tools (like patcher?)

### DIFF
--- a/patchwork-api-base/src/main/java/net/patchworkmc/annotations/Stubbed.java
+++ b/patchwork-api-base/src/main/java/net/patchworkmc/annotations/Stubbed.java
@@ -1,0 +1,33 @@
+/*
+ * Minecraft Forge, Patchwork Project
+ * Copyright (c) 2016-2020, 2019-2020
+ *
+ * This library is free software; you can redistribute it and/or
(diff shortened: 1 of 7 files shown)


famous1622:
AFAICT All functionality that these classes had before is still present. Maybe add a comment at the top of these classes explaining not to try to use them in other modules, part of the appeal of codegen for me was that it made it pretty clear that the classes weren't meant for us.

kitlith:
huh, i thought they all ready had comments at the top. at least a couple of them do. I'll add the comments.

I'm still open to switching back to codegen if everyone else prefers codegen, but my personal scales have been tipped away from it and the other people I've asked have been non-committal, so I just. made a decision.

TheGlitch76 (approved):
concept looks good, will merge later

PR #204, opened 2020-11-19
God class codegen redux (now with an additional hack!)

kitlith:
#202 but it's a rewrite using javapoet and a hack to run an annotation processor on multiple projects at once:

https://github.com/kitlith/patchwork-api/blob/b91a4ebe7a2099fdbde457d953643d428249269e/patchwork-god-classes/build.gradle#L37-L68

I like this much better than #202, but I'm not sure how I feel about the hack involved. Given the usage of javapoet and how well it integrates with annotation processors, I feel a lot better about this version probably not breaking at some point, and being easier to extend/fix.

I'm leaving #202 open for now in the case that we want codegen, but don't want to use the hack in here.

... it appears that you can serialize individual methods and fields w/ javapoet, so if we don't like the hack, I can generate fragments of code which I then piece together later, and I can even reuse most of the code I wrote for this version. yay javapoet.

First 3 commits are from #201, please go there if you have concerns about how that's laid out.


kitlith:
something something we can vote on if we want to use the hack or not?

:+1: for 'LGTM'
:-1: for 'god why would you do this'

PR #205, opened 2020-11-21
Runtime remapping support

cittyinthecloud:
This PR is a draft for remapping reflection accesses to MC classes from SRG to whatever mappings are used at runtime. My goals are

- [x] ObfuscationReflectionHelper remaps to runtime mappings
~~- [ ] ReflectionHelper remaps if it's handed SRG names~~
- [ ] (Potentially) have Patcher wrap Java's reflection API to static methods in API (would require a transformer over there)


famous1622:
It's been a while since I used Forge, ReflectionHelper is no longer a thing

PR #206, opened 2020-11-26
Make methods needed for Iron Chests accessible

keplersj:
PR makes two methods needed for Iron Chests to launch accessible.

Screenshots:

Happy to change location of Access Widener rulesl


WalkerKnapp (changes requested):
These might be better in `patchwork-gui`, in my opinion. An access widener would need to be created there, though.

TheGlitch76:
This is fine as-is for now--gui module doesn't really handle this currently and stuff in patchwork-fml is meant to be sorted out as it comes along

WalkerKnapp (approved):
Sounds good then 👍

TheGlitch76 (approved):
Thanks for this!

PR #209, opened 2020-12-03
Changes for Neat

stairman06:
Adds RenderWorldLastEvent and an AW for `Entity.getSavedEntityId()`. These changes are required for [Neat](https://www.curseforge.com/minecraft/mc-mods/neat). 

Note that Neat does not work when patched with upstream Patcher, as there is a bug where SRG constant strings are not remapped. I have [a fork](https://github.com/stairman06/patchwork-patcher/tree/stringconstant-hack) (`stringconstant-hack` branch) which re-adds the remapping functionality. I can PR to Patcher if that is what I should do, however there is probably a better way to solve the problem than what I did in my fork.


TheGlitch76 (approved):
LGTM! Sorry for the wait!

PR #214, opened 2020-12-21
Hand Render Events

stairman06:
RenderHandEvent and RenderSpecificHandEvent. Not sure if the `@ModifyConstant` for the if statement is the best way, but it works.


kitlith:
The `@ModifyConstant` is fine, especially since you documented what is going on with it. Sometimes, the cleanest thing we can do is something like that in these cases.

I'd just double check that there are no other instances of '0' in the method, so that unintentional changes are not made.

stairman06:
Thanks. I've double checked and there were other 0s in the method, so I've added an ordinal to the Constant.

TheGlitch76 (approved):
lgtm

PR #217, opened 2020-12-23
ItemGroup#getGroupCountSafe

stairman06:
Tiny method. One step closer to loading [Create](https://www.curseforge.com/minecraft/mc-mods/create).


stairman06:
Closing this for now, it can be done when PatchworkItemGroup is updated to 1.16.

PR #218, opened 2020-12-26
Update events-input

YTG1234:
The only real change was `getKeyboard()`.

I can't test this as not all the required modules are updated, but it's a simple update so it should be fine.


patchwork-events-input/src/main/java/net/patchworkmc/mixin/event/input/MixinKeyboard.java
```diff
@@ -33,11 +34,12 @@
 @Mixin(Keyboard.class)
 public abstract class MixinKeyboard {
 	@Shadow
-	MinecraftClient client;
+	@Final
+	private MinecraftClient client;
 
 	@Inject(method = "onKey", at = @At("RETURN"))
```
TheGlitch76, line 41: Use an ordinal, change to TAIL, or comment you're specifically targeting all return points.

TheGlitch76 (changes requested):

TheGlitch76 (approved):

PR #221, opened 2020-12-31
Update patchwork-events-rendering

stairman06:
Ports patchwork-events-rendering to 1.16. This PR also updates the Fabric API requirement to 0.29.0 as it uses the new WorldRenderEvents. 

This also brings back `ModLoader` in patchwork-fml, so I can call `postEvent`, I don't think that should break anything though. The actual implementation of `postEvent` here doesn't actually match the forge implementation ([see here](https://github.com/PatchworkMC/YarnForge/blob/ccd6dc9c99d34bcec90f79eb67019fdf6ac05ee8/src/main/java/net/minecraftforge/fml/ModLoader.java#L321)), because I thought implementing IModBusEvent would be too out of scope for this PR.


stairman06:
Don't merge this yet! I've encountered a strange error when doing some testing and I want to verify that it's not this PR causing the problem.

stairman06:
After some local testing I'm unable to fix the issue. Either way, I think the issue being because of this PR is unlikely - all the events fire at the right spot. There is probably something else I'm missing, or Forge is patching something else that I've missed. Anyway, I think this PR is good for now, it implements all the previous events.

PR #224, opened 2021-01-09
Update patchwork-extensions and patchwork-extensions-entity

WalkerKnapp:
It's finally done! This took way longer to test than it did to implement, but for the most part, both are now updated.

Some TODO items for the future:
- Once patchwork-extensions-block is implemented, `BlockHarvestManager#onFarmlandTrample` needs to be reconciled with `IForgeEntity#canTrample`
- When patchwork-capabilities makes `Entity` a `CapabilityProviderHolder`, `IForgeEntity#revive` should call `CapabilityProviderHolder#reviveCaps`


TheGlitch76:
Hi yes, can you please annotate any method in IForge* that hasn't had all of its callsites implemented with @Stubbed?
It is used in AreWePatchworkYet and may be detected in a future version of Patcher

TheGlitch76 (approved):
didn't see any big issues from a cursory glance, going to call it good enough for the sake of porting
lgtm

PR #226, opened 2021-01-11
Update world events

florensie:
This is still untested, but open for reviews.
- Only updated existing patchwork events, no new ones were added
- `WorldTypeEvent` was removed upstream
- `DifficultyChangeEvent` implementation was broken upstream, but now fixed and implemented here as well


patchwork-events-world/src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -159,35 +158,25 @@ public BlockPos getPos() {
...
 
-		public LevelInfo getSettings() {
+		public ServerWorldProperties getSettings() {
 			return settings;
 		}
-
-		@Override
-		public boolean isCancelable() {
-			return true;
-		}
```
WalkerKnapp: This event is still cancellable in forge, this should be readded.

patchwork-events-world/src/main/java/net/minecraftforge/event/world/WorldEvent.java
```diff
@@ -159,35 +158,25 @@ public BlockPos getPos() {
...
 
-		public List<SpawnEntry> getList() {
+		public List<SpawnSettings.SpawnEntry> getList() {
 			return list;
 		}
-
-		@Override
-		public boolean isCancelable() {
-			return true;
-		}
```
WalkerKnapp: This event is still cancellable in forge, this should be readded.

WalkerKnapp (changes requested):
I just wrote up a test mod, everything works well.  PotentialSpawns and CreateSpawnPosition just need to be marked cancellable.

patchwork-events-world/src/main/java/net/minecraftforge/event/world/WorldEvent.java
florensie: Don't know how I missed that.

TheGlitch76 (approved):

WalkerKnapp (approved):

